PR #13, opened 2021-04-28
Adding new meal prompt

cheburan:
Adding new meal prompt (survey_client & admin_client)


mechkg:
Merged manually due to massive amount of conflicts

PR #88, opened 2022-10-03
fix(survey): V4-427 (bug: 477/508): DrinkScale Overlay Image sliding

cheburan:
DrinkScale prompt Overlay image slider.


cheburan:
DrinkScale prompt (V4-427, 537, bug: 477/508)

PR #153, opened 2023-07-21
fix(developer): modify the developer resources documentation.

mucc001:


docs/developer/getting-started.md
```diff
@@ -5,6 +5,7 @@
 Steps to getting a local instance running
 
 - Run Dev VM (download from S3 bucket - contact the Intake24 team), which maps out the database on 192.168.56.10:5432 (PostgreSQL). Start this VM.
+- Alternatively, obtain database snapshots from the intake24 team and import them into a database browser (e.g. DBeaver). Please check the guidance on [importing database snapshots with DBeaver](https://dbeaver.com/docs/wiki/Backup-Restore/).
```
lukashroch, line 8: DBeaver is the database browser. Rephrase to `database (e.g. using DBeaver)`?

docs/developer/getting-started.md
```diff
@@ -24,7 +26,8 @@ Follow the readme instructions in the root of the repository.
 
 ## Databases
 
-Standalone database snapshots are currently unavailable, please use the [development virtual machine](/developer/vm.html).
+Please contact the intake24 team for the latest development virtual machine. [Please also check the guidance on using the development virtual machine](https://docs.intake24.org/developer/vm.html).
+Alternatively, obtain database snapshots from the intake24 team and import them into a database browser (e.g. DBeaver). Please check the guidance on [importing database snapshots with DBeaver](https://dbeaver.com/docs/wiki/Backup-Restore/). If using the database snapshots, you need to set up your own Redis instance.
```
lukashroch, line 30: Redis is needed regardless whether you use DB snapshots. I think last sentence can be droppped.

docs/developer/getting-started.md
```diff
@@ -47,11 +50,11 @@ Copy the `.env-template` file to `.env` and change the following settings:
...
+:::
 
-The default user with superuser rights is `admin` and the password is `intake24`. Refer to the [API docs](/api/) for
-further details.
+## API access
 
-::: tip
-Use the e-mail log in endpoint (`/api/login`) for the admin user even though "admin" is not a valid e-mail address.
-:::
+Please add your email in DBeaver `users` tables, then go to `apps/cli`, and run `pnpm cli:dev hash-password yourNewPassword`.
```
lukashroch, line 59: I would extend this a bit that user needs to create `user` and `user_password` records in database manually. Then add the password hash using the cli. This one only worked for you as I gave you snapshot where you had your accoult already created.

This DB setup bit is still a bit combersome as we don't have a clean way, it always depends if it's created from clean DB or some snapshot etc. I think we should have for this at least a CLI command that creates whole new account without this manual hassle to streamline this a bit.

lukashroch (commented):
Couple of comments I would modify, but otherwise looks good!

PR #159, opened 2023-08-08
feat(bottom navigation): streamline bottom navigation

mucc001:
Make sure that the review button will always appear as the leftmost option on the mobile bottom navigation bar. 
https://intake24.atlassian.net/browse/V4-830


lukashroch:
I'll close this as superseeded by #162

PR #166, opened 2023-08-14
fix(survey): handle associated foods with sab

lukashroch:
Handle associated foods with same as before

- prevents food to be saved as SAB prematurly when assoc food PSMs are not finished yet
- convert assoc food propery to flag
- trigger SAB check on psm finish or assoc prompt finish for no linked food cases


mechkg (approved):
This looks fine to me, the dedicated associated foods complete field looked a bit out of place anyway.

PR #179, opened 2023-09-14
feat(quicklist): new final review list with checkbox and scroll options

mucc001:
Create reviewed-list.vue which adds a Reviewed checkbox after each meal in the original quick list. Add a new array reviewed to store meal that is already reviewed.


apps/admin/src/components/prompts/standard/submit-prompt.vue
```diff
@@ -1,16 +1,74 @@
 <template>
-  <v-tab-item key="options" value="options"></v-tab-item>
+  <v-tab-item key="options" value="options">
+    <v-card-title>"Final Review Options"</v-card-title>
```
lukashroch, line 3: Can you please move hardcoded texts to translations?

apps/admin/src/components/prompts/standard/submit-prompt.vue
```diff
@@ -1,16 +1,74 @@
...
+
 import { basePrompt } from '../partials';
 
 export default defineComponent({
   name: 'SubmitPrompt',
 
   mixins: [basePrompt],
+
+  props: {
+    desktopReview: {
```
lukashroch, line 33: This are now redundant

apps/admin/src/components/prompts/standard/submit-prompt.vue
```diff
@@ -1,16 +1,74 @@
...
+      type: [Boolean, String] as PropType<Prompts['submit-prompt']['desktopReview']>,
+      required: true,
+    },
+    mobileReview: {
```
lukashroch, line 37: This are now redundant

apps/admin/src/components/prompts/standard/submit-prompt.vue
```diff
@@ -1,16 +1,74 @@
...
+    review: {
+      handler(newVal) {
+        this.localReview = { ...newVal };
+      },
+      immediate: true,
+    },
+  },
+
+  methods: {
+    updateDesktopOptions(value: string | boolean) {
```
lukashroch, line 62: This are now redundant

apps/admin/src/components/prompts/standard/submit-prompt.vue
```diff
@@ -1,16 +1,74 @@
...
+      required: true,
+    },
+    review: {
+      type: Object as PropType<Prompts['submit-prompt']['review']>,
+    },
+  },
+
+  data() {
+    return {
+      localReview: { ...this.review },
```
lukashroch, line 48: Istead of keeping local state, you could do either the option we discussed or use a computed with setter
```
computed({
get() {
return this.review;
}
set(value) {
this.update('review', value);
}
})
```
I think that should work, watchers tend to complicate it

apps/survey/src/components/prompts/standard/SubmitPrompt.vue
```diff
@@ -13,31 +41,193 @@
...
+
   setup(props, ctx) {
     const { action } = usePromptUtils(props, ctx);
 
-    const isValid = true;
+    const reviewed = ref<string[]>([]);
+
+    const reviewedMobile = ref<string[]>([]);
+
+    const isValid = ref<boolean>(false);
```
lukashroch, line 100: you can remove `boolean`, also from few more below, its implied

apps/survey/src/components/prompts/standard/SubmitPrompt.vue
```diff
@@ -1,7 +1,35 @@
 <template>
   <card-layout v-bind="{ food, meal, prompt, section, isValid }" @action="action">
+    <review-meal-list
```
lukashroch, line 3: use <component :is="myComponentNameValue"> to selelct component based on conditions, lots of repetition.

Do we also need component for each layout or behaviour logic? I think we said that on mobile it practically the same? If it cannot fit to less components, we should extract the common logic at least.

packages/common/src/prompts/prompts.ts
```diff
@@ -215,7 +215,12 @@ export type Prompts = {
   'review-confirm-prompt': BasePrompt & { component: 'review-confirm-prompt' };
   'same-as-before-prompt': BasePrompt & { component: 'same-as-before-prompt' };
   'split-food-prompt': BasePrompt & { component: 'split-food-prompt' };
-  'submit-prompt': BasePrompt & { component: 'submit-prompt' };
+  'submit-prompt': BasePrompt & {
+    component: 'submit-prompt';
+    desktopReview: false | 'scroll' | 'checkbox';
```
lukashroch, line 220: redundant

packages/common/src/prompts/prompts.ts
```diff
@@ -215,7 +215,12 @@ export type Prompts = {
...
+    mobileReview: false | 'scroll' | 'checkbox';
```
lukashroch, line 221: redundant

apps/survey/src/components/prompts/standard/SubmitPrompt.vue
```diff
@@ -13,31 +41,193 @@
...
+  props: {
+    meals: {
+      type: Array as PropType<MealState[]>,
+      required: true,
+    },
+    prompt: {
+      type: Object as PropType<Prompt>,
+      required: true,
+    },
+    review: {
```
lukashroch, line 88: redundant? this comes from `props.prompt`

apps/survey/src/components/prompts/standard/SubmitPrompt.vue
```diff
@@ -13,31 +41,193 @@
...
+  },
+
   mixins: [createBasePrompt<'submit-prompt'>()],
 
+  props: {
+    meals: {
+      type: Array as PropType<MealState[]>,
+      required: true,
+    },
+    prompt: {
```
lukashroch, line 84: already added in `createBasePrompt` and should be typed as `submit-prompt`

apps/survey/src/components/prompts/standard/SubmitPrompt.vue
```diff
@@ -13,31 +41,193 @@
...
+
+    const isValidMobile = ref<boolean>(false);
+
+    const reachedBottom = ref<boolean>(false);
+
+    const reachedBottomMobile = ref<boolean>(false);
+
+    const showReviewMealListCheckbox = computed(() => {
+      if (
+        props.prompt.component === 'submit-prompt' &&
```
lukashroch, line 110: `props.prompt.component === 'submit-prompt'` this check won't be needed, if you remove props.prompt property, it will be typed as submit-prompt. Same for few more checks below, which can then be simplified

apps/survey/src/components/prompts/standard/SubmitPrompt.vue
```diff
@@ -13,31 +41,193 @@
...
+    const setReachedBottom = (status: boolean) => {
+      reachedBottom.value = status;
+    };
+
+    const setReachedBottomMobile = (status: boolean) => {
+      reachedBottomMobile.value = status;
+    };
+
+    watch(
+      [
```
lukashroch, line 169: wonder if we can simplify this, this will be expensive watcher

lukashroch (changes requested):

apps/admin/src/components/prompts/standard/submit-prompt.vue
mucc001: Thanks for the suggestion. I have changed it to the option we discussed. 
```
<v-combobox
        v-for="key in Object.keys(review)"
        :key="key"
        :label="key"
        :value="review[key]"
        :items="[false, 'checkbox', 'scroll']"
        @change="update('review', {...review, [key]: $event})"
>
```

apps/survey/src/components/prompts/standard/SubmitPrompt.vue
```diff
@@ -13,31 +41,159 @@
...
-    const isValid = true;
+    const reviewed = ref<string[]>([]);
+
+    const reviewedMobile = ref<string[]>([]);
+
+    const reachedBottom = ref<boolean>(false);
+
+    const reachedBottomMobile = ref<boolean>(false);
+
+    const showReviewMealListCheckbox = computed(() => {
```
lukashroch, line 96: You can also inline these few checks below to simple expressions now:
```ts
const showReviewMealListCheckbox = computed(() => props.prompt.review['desktop'] === 'checkbox');
```

packages/common/src/prompts/standard.ts
```diff
@@ -130,6 +130,12 @@ export const submitPrompt: Prompts['submit-prompt'] = copy({
   type: 'standard',
   id: 'submit-prompt',
   name: 'Submit prompt',
+  desktopReview: false,
```
lukashroch, line 133: to remove L133-134

apps/survey/src/components/prompts/standard/SubmitPrompt.vue
mucc001: Thanks for the suggestion. The watcher is not necessary and I have changed it to a few computed properties.

apps/survey/src/components/prompts/standard/SubmitPrompt.vue
mucc001: Thanks. I have inlined these checks.

apps/survey/src/components/prompts/standard/SubmitPrompt.vue
mucc001: I have kept only two final review list components—one for the desktop and the other for the mobile versions. I use v-if and some flags to switch between the checkbox and the scroll bar options.

apps/admin/src/components/prompts/standard/submit-prompt.vue
```diff
@@ -1,16 +1,41 @@
 <template>
-  <v-tab-item key="options" value="options"></v-tab-item>
+  <v-tab-item key="options" value="options">
+    <v-card-title>{{ $t('survey-schemes.prompts.submit-prompt.optionTitle') }}</v-card-title>
```
lukashroch, line 3: rename the key to something like `review` ? ...There will be more an more options once more features implemented. Maybe one title is enough, see options in other prompt.

apps/admin/src/components/prompts/standard/submit-prompt.vue
```diff
@@ -1,16 +1,41 @@
...
+    <v-card-subtitle>{{
+      $t('survey-schemes.prompts.submit-prompt.optionSubtitle')
+    }}</v-card-subtitle>
+    <v-card-text>
+      <v-combobox
+        v-for="key in Object.keys(review)"
+        :key="key"
+        :items="[false, 'checkbox', 'scroll']"
+        :label="key"
```
lukashroch, line 12: `:label="survey-schemes.actions.layouts.${key}"`

apps/survey/src/components/prompts/createBasePrompt.ts
```diff
@@ -26,9 +26,46 @@ export default <P extends keyof Prompts, F extends FoodState = EncodedFood>() =>
         type: Object as PropType<Prompts[P]>,
         required: true,
       },
+      commonPrompt: {
```
lukashroch, line 29: is this used?

apps/survey/src/components/prompts/createBasePrompt.ts
```diff
@@ -26,9 +26,46 @@ export default <P extends keyof Prompts, F extends FoodState = EncodedFood>() =>
...
+        type: Object as PropType<Prompt>,
+        required: false,
+      },
       section: {
         type: String as PropType<PromptSection>,
         required: true,
       },
     },
+
```
lukashroch, line 38: is this also used, whole setup below?

lukashroch (changes requested):

apps/admin/src/components/prompts/standard/submit-prompt.vue
mucc001: Thanks. It seems that we need to add $t() around it `:label="$t(`survey-schemes.actions.layouts.${key}`)"`

PR #191, opened 2023-10-18
feat(api): Adding Survey Slug to the Survey Submission output

cheburan:
A small change to the survey submission is adding a Survey Slug as a parameter to the output. For the sake of the external modules (a.k.a. intake24-dietician).

### DIFF
--- a/packages/i18n/src/admin/en/survey-schemes.json
+++ b/packages/i18n/src/admin/en/survey-schemes.json
@@ -144,7 +144,6 @@
       "user": "User record fields",
       "userCustom": "User custom fields",
       "survey": "Survey record fields",
-      "surveyCustom": "Survey custom fields",
       "submission": "Submission fields",
       "submissionCustom": "Submission custom fields",
       "meal": "Meal record fields",
--- a/packages/i18n/src/admin/en/survey-schemes.json
+++ b/packages/i18n/src/admin/en/survey-schemes.json
@@ -144,6 +144,7 @@
       "user": "User record fields",
       "userCustom": "User custom fields",
       "survey": "Survey record fields",
+      "surveyCustom": "Survey custom fields",
       "submission": "Submission fields",
       "submissionCustom": "Submission custom fields",
       "meal": "Meal record fields",
--- a/apps/api/src/services/admin/data-export/data-export-fields.ts
+++ b/apps/api/src/services/admin/data-export/data-export-fields.ts
@@ -120,6 +120,11 @@ const dataExportFields = () => {
         return aliases && aliases.length ? aliases[0].username : undefined;
       },
     },
+    {
+      id: 'slug',
+      label: 'Survey Slug',
+      value: ({ food }: ExportRow) => food.meal?.submission?.survey?.slug,
+    },
   ];
 
   /**
--- a/apps/api/src/services/admin/data-export/data-export.service.ts
+++ b/apps/api/src/services/admin/data-export/data-export.service.ts
@@ -16,7 +16,6 @@ import type {
   WhereOptions,
 } from '@intake24/db';
 import { NotFoundError } from '@intake24/api/http/errors';
-import { logger } from '@intake24/common-backend/services';
 import {
   Op,
   Survey,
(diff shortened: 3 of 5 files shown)


lukashroch:
This will only add it to json response output. I think we should include it as field to data export. Maybe add survey name as well?

lukashroch (commented):
This will only add it to json response output. I think we should include it as field to data export. Maybe add survey name as well?

cheburan:
@lukashroch Reasonable.  Adding a field object to `data-export-files.ts` will do the trick, as I understand. 

Will do that once I figure out why my data export doesn't work and can't be configured via the admin dashboard.
Have some issues with modifying the data export for the survey schema. I don't think it is connected to the task itself, rather then the survey schema. For some reason, I can't save the changes in the exported fields (admin dashboard) list even if I don't change the code and list of fields. Getting the following error: 
```
ExportSections/3/id must be equal to one of the allowed values. Where 3 is `surveyCustom`. 

Not sure if it is only me, but it is missing the translation in the admin dashboard, too - `survey-schemes.data-export.sections.surveyCustom.`
```
when sending the request to `http://{HOST}api/admin/survey-schemes/8`.

It seems connected to the `surveyCustom` section in the Exported Fields. Similar to the `TypeError: dataExportMapper[section.id] is not a function` error I get when exporting the data. In the `getExportFields` function on the execution of `dataExportMapper[section.id](section.fields)` on the `surveyCustoms` section (API). 

Do you know if surveyCustom was added recently? Do I need to run any migration/code or change the survey schema?

lukashroch:
It was added couple of months ago. I thought admin screen is merging with default so new section would me merged. We'll probably have to look into that.
If you paste in into raw json
```
[
  { id: 'user', fields: [] },
  { id: 'userCustom', fields: [] },
  { id: 'survey', fields: [] },
  { id: 'submission', fields: [] },
  { id: 'submissionCustom', fields: [] },
  { id: 'meal', fields: [] },
  { id: 'mealCustom', fields: [] },
  { id: 'food', fields: [] },
  { id: 'foodCustom', fields: [] },
  { id: 'foodFields', fields: [] },
  { id: 'foodNutrients', fields: [] },
  { id: 'portionSizes', fields: [] },
]
```

I think that should work?

cheburan:
> It was added couple of months ago. I thought admin screen is merging with default so new section would be merged. We'll probably have to look into that. If you paste in into raw json
> 
> ```
> [
>   { id: 'user', fields: [] },
>   { id: 'userCustom', fields: [] },
>   { id: 'survey', fields: [] },
>   { id: 'submission', fields: [] },
>   { id: 'submissionCustom', fields: [] },
>   { id: 'meal', fields: [] },
>   { id: 'mealCustom', fields: [] },
>   { id: 'food', fields: [] },
>   { id: 'foodCustom', fields: [] },
>   { id: 'foodFields', fields: [] },
>   { id: 'foodNutrients', fields: [] },
>   { id: 'portionSizes', fields: [] },
> ]
> ```
> 
> I think that should work?

Yep, it all worked. Now, Survey Slug is added as an option for the Survey fields' export 
Screenshot 2023-10-20 at 10 22 27

lukashroch:
Sorry,  `surveyCustom` actually became `submissionCustom` and `survey` is referencing the survey rather than the `submission`

packages/i18n/src/admin/en/survey-schemes.json
```diff
@@ -144,6 +144,7 @@
       "user": "User record fields",
       "userCustom": "User custom fields",
       "survey": "Survey record fields",
+      "surveyCustom": "Survey custom fields",
```
lukashroch, line 147: Not used anymore, remove please.

lukashroch (changes requested):
Please add it also to docs: https://github.com/MRC-Epid-it24/intake24/blob/master/docs/admin/surveys/schemes.md#survey-record-fields

cheburan:
> Please add it also to docs: https://github.com/MRC-Epid-it24/intake24/blob/master/docs/admin/surveys/schemes.md#survey-record-fields

added

PR #204, opened 2023-11-17
fix(prompt): clean stale local prompt states (V4-984)

lukashroch:
- removes out-of-sync local prompt states during food change/deletion

This seems to happen in various cases:

At the moment local prompt stores are usually cleaned once answer is committed. This is not the case at the moment for PSM stores (I think this was allowed to allow go back functionality when progressing in PSM).

However, user can take action in middle of the PSM, which has not been commit yet, which can result in invalid state due to stale local prompt store (which is then out of sync with survey state)

1) user is middle of PSM and changes the food to something else -> local PSM store if kept from original. Can result in discrepancies-only or also in error - e.g. when there is invalid standard unit not anymore on new food
2) deleting food in middle of PSM
3) deleting meal
4) can probably also happen with `back` button, if user deletes/changes foods, tries to go back, and local prompt invalid state is loaded.

### DIFF
--- a/apps/survey/src/stores/prompt.ts
+++ b/apps/survey/src/stores/prompt.ts
@@ -46,12 +46,20 @@ export function getOrCreatePromptStateStore<T extends object>(
             [foodOrMealId]: { ...this.prompts[foodOrMealId], [promptId]: data },
           };
         },
-        clearState(foodOrMealId: string, promptId: string) {
-          if (this.prompts[foodOrMealId]?.[promptId])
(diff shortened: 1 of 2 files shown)


mechkg:
This is a much needed fix, and the implementation looks good to me. There's probably still going to be some weird corner cases but I guess we should just go ahead and let people test it.

mechkg (approved):
This is a much needed fix, and the implementation looks good to me. There's probably still going to be some weird corner cases but I guess we should just go ahead and let people test it.

PR #220, opened 2024-01-07
Importer option selections: Specific step/module execuition

cheburan:
Added option to execute a specific step of `import-package` command by adding `-m` option.

option '-m, --modules-for-execution [modules-for-execution-option]' 
List of arguments/options: 
 - 'enabled-local-foods' 
 - images-as-served, 
 - locales, 
 - nutrients, 
 - global-foods, 
 - local-foods, 
 - enabled-local-foods, 
 - all **[Default if not chosen otherwise]**.

### DIFF
--- a/apps/cli/src/commands/packager/importer-v4.ts
+++ b/apps/cli/src/commands/packager/importer-v4.ts
@@ -1,4 +1,3 @@
-import { log } from 'node:console';
 import os from 'node:os';
 
 import decompress from 'decompress';
@@ -39,12 +38,16 @@ export type ImporterSpecificModulesExecutionStrategy =
 
 export interface ImporterOptions {
   onConflict?: ConflictResolutionStrategy;
-  modulesForExecution?: ImporterSpecificModulesExecutionStrategy;
+  modulesForExecution?: ImporterSpecificModulesExecutionStrategy[];
 }
+
+export type availableModules = {
+  [key in ImporterSpecificModulesExecutionStrategy]: () => Promise<void>;
+};
 
 const defaultOptions: ImporterOptions = {
   onConflict: 'abort',
-  modulesForExecution: 'all',
+  modulesForExecution: ['all'],
 };
 
 export class ImporterV4 {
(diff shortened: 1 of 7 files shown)


apps/cli/src/commands/packager/importer-v4.ts
```diff
@@ -485,17 +498,44 @@ export class ImporterV4 {
...
+            await this.importAsServedSets();
+          } finally {
+            await this.cleanUpPackage();
+          }
+          break;
+        default:
+          throw new Error(
+            `Unexpected module for execution option: ${this.options.modulesForExecution}`
+          );
+      }
```
mechkg: Hi @cheburan, this switch statement looks suspicious -- "modulesForExecution" is probably an array an you're doing a switch on it. Does this work? 😅

cheburan: HI @mechkg , That code you highlighted doesn't look final; it changed since. I added more options to it. 

But the idea was that the users could choose either one specific module or "all", hence the switch that only allows them to choose either "all" or one of the modules (locale, images-as-served, ...). It doesn't allow an array of options atm. It's better to prevent users from executing a random combination of steps/modules. Should CLI allow it?

lukashroch: I'll rebase it, think it's a bit muddled with merged stuff from main

mechkg: @cheburan I see, but it needs to allow multiple modules because sometimes one is dependent on the other (local-foods needs as-served and nutrient-tables). These dependencies don't need to be enforced, but allowing several options to be executed in the specified order should be allowed  (e.g. --as-served --nutrient-tables --local-foods)

cheburan: Ok true. Added the ability to execute multiple steps in a user-specified order. Like: 
```
pnpm run cli import-package -c overwrite -m locales nutrients v4 /path/to/the/archived/ARCHIVED_FOLDER.zip
```
all based on this: 
[https://github.com/MRC-Epid-it24/intake24/pull/220/files#diff-d43093ebd3ed9800863ee697acfb479d16d6efc76928a54e2469e0ec3b9c7267R93](url)

PR #222, opened 2024-01-10
feat(submission): collect parent recipe food record (V4-1013)

lukashroch:
Collects parent recipe food record of the child ingredients
- allows ingredients to be linked to the recipe through parent
- collects recipe information on parent record


cheburan:
@lukashroch, all look reasonable, and the data export is pretty clear, too.

PR #226, opened 2024-01-17
Sandwich builder fixes v4 807

cheburan:
- [x] Drop the ‘browse all foods’ sentence and instead replace it with '<StepName> categories as per sys v3

- [x] Admin: Add “Required” Option to the Admin Dashboard (locale → recipe foods → steps)

- [x] Survey: Make some steps Required (configurable via locale->recipe foods → steps

- [x] Survey: Change text for the Missing Food Prompt 

- [x] ‘I DIDN'T HAVE ANY X' for all steps won’t be possible to continue, and the system will show the message that at least one ingredient (step) is needed

### DIFF
--- a/apps/admin/src/views/locales/recipe-foods/steps-dialog.vue
+++ b/apps/admin/src/views/locales/recipe-foods/steps-dialog.vue
@@ -213,8 +213,6 @@ export default defineComponent({
           return item;
         });
 
-      console.log(form.items);
-
       const items = await form.post<LocaleRecipeFoodSteps[]>(
         `admin/locales/${props.locale.id}/recipe-foods/${props.activeRecipeFoodId}/steps`
       );
--- a/apps/admin/src/views/locales/recipe-foods/steps-dialog.vue
+++ b/apps/admin/src/views/locales/recipe-foods/steps-dialog.vue
@@ -96,6 +96,12 @@
                     :label="$t('locales.recipe-foods.repeat')"
                     name="repeatable"
                   ></v-switch>
+                  <v-switch
+                    v-model="item.required"
+                    hide-details="auto"
+                    :label="$t('locales.recipe-foods.require')"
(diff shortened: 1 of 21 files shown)


cheburan:
Will need to apply migrations.

apps/survey/src/components/prompts/portion/RecipeBuilderPrompt.vue
```diff
@@ -131,10 +140,12 @@ export default defineComponent({
 
   setup() {
     const { translate } = useI18n();
+    const missingAllIngridientsDialog = ref(true);
```
lukashroch, line 143: Looks like this one is not used?

apps/api/src/services/admin/locale.service.ts
```diff
@@ -9,6 +11,7 @@ import type {
 } from '@intake24/common/types/http/admin';
 import { NotFoundError } from '@intake24/api/http/errors';
 import { addDollarSign } from '@intake24/api/util';
+import { logger } from '@intake24/common-backend/services';
```
lukashroch, line 14: not used

apps/api/src/services/admin/locale.service.ts
```diff
@@ -1,3 +1,5 @@
+import { log } from 'console';
```
lukashroch, line 1: not used

apps/admin/src/views/locales/recipe-foods/steps-dialog.vue
```diff
@@ -207,6 +213,8 @@ export default defineComponent({
           return item;
         });
 
+      console.log(form.items);
```
lukashroch, line 216: redundant

lukashroch (changes requested):
looks good, just couple of unused items to remove.

PR #230, opened 2024-01-27
Sliding scales rework

mechkg:
@lukashroch Could you please have a look at this change:

https://github.com/MRC-Epid-it24/intake24/commit/4fd8c3f6aed755a0f0efa2e763f88849aba8075c

The "Repo - unit tests" step was failing when ran from the repository root directory, which probably had to do with the jest module resolution config but instead of trying to fix it I though that it made more sense to run the unit tests in the apps directories similar to how integration tests are run.

This change works on my machine but when run by github it says "No projects matched the filters in "/home/runner/work/intake24/intake24""

I was wondering if the "Repo unit tests" task should rather be split and ran specifically for each app if required? It looks a bit out of place in the API checks section.


lukashroch:
I had the same issue with integration tests:

command used to be:
```
test:integration": "pnpm --filter * test:integration
```

which works fine locally for me, but not on GH actions for whatever reason ...

Couldn't figure out why it didn't match, (maybe a bug in pnpm and some OS-cli nuance?) ... so swapped to globs for now
```
pnpm --filter ./apps/{api,portal} test:integration
```
You could do similar with globs
```
pnpm --filter ./{apps,packages}/**/ test:unit
```

But you will have to install and configure jest in each package as it has no test suite setup on package-level. Original idea was not having to configure test suite for each package individually and run unit tests from top level with same config. True unit tests should not really require much ext deps, so was hoping it could be simplified for whole repo?
... a bit harder for integrations tests as they require to spin up more ext deps depending what's being tested ... so each config is unique. `test:unit` in `api` is kind of just a convenience/leftover since it already had integration tests set up. So rest of apps/packages in the repo would need to be set up with test suite.

I can have a look at the types, hopefully shouldn't be difficult to track down?

lukashroch:
OK, those types should be fixed.

All packages export public code that can be used to top level, so they also need to be imported only from top-level. Or at least that was the idea to keep each package functionality a bit more self-contained. There are two exceptions to that, `common` and `common-backup`, which is a mix of types and various utils, so there are named conflicts ... so not possible at current state.

packages/db/src/kysely-database.ts
```diff
@@ -0,0 +1,104 @@
...
+    this.dbLogger = logger.child({ service: 'Kysely DB' });
+  }
+
+  private configKyselyDialect(database: DatabaseType): Dialect {
+    // Postgres hard-coded for experimental Kysely code, in case we decide to keep using it
+    // return the correct dialect type based on the database config
+
+    return new PostgresDialect({
+      pool: new Pool({
+        database: this.config[this.env][database].database,
```
lukashroch, line 44: Could you please add connection string option for kysely db? It's already in db config (`this.config[this.env][database].url`) for current setup and used in deployments.

apps/api/.env-template
```diff
@@ -133,3 +133,14 @@ CAPTCHA_SECRET=
 WEBPUSH_SUBJECT=mailto:example@domain.com
 WEBPUSH_PUBLIC_KEY=
 WEBPUSH_PRIVATE_KEY=
+
+# Image processing
```
lukashroch, line 137: Could you include new configs to docs? https://docs.intake24.org/config/api/

Maybe drop `APP_` prefix for consistently, as it's a different part of the config?

PR #235, opened 2024-02-06
V4 1016 food index rebuild on changes

cheburan:
- Added foodControllers for rebuilding all indexes and specified indexes only (using existing NodeJs Workers)
- Added Jobs for rebuilding All and Specified only indexes  (using existing NodeJs Workers and 2 new Redis Jobs)
- **Added Redis Set as a message buffer for the Scheduled Jobs to read from **
- Added triggers for Specified Indexing Job (Admin Locales changes in Synonyms and Recipe Foods) to Check the Redis Set and executed reindexing for the "modified' locales
- Added Redis Pub: Sub as a message broker between the Job and the food-indexing processes running in the cluster


cheburan:
Redis Stream: Adding messages is working, but the format of messages during the read() is weird. Doesn't work properly.

apps/api/src/config/set.ts
lukashroch, line 1: Do we still need this, if we have pub/sub. This looks like more of temp value, so could be easily used in cache redis client? set ability can be moved there is it's needed.

apps/api/src/http/controllers/food-search.controller.ts
```diff
@@ -62,6 +77,26 @@ const foodSearchController = ({
     };
   }
 
+  // API endpoint for rebuilding the food index
+  const rebuildFoodIndex = async (req: Request, res: Response): Promise<void> => {
+    logger.debug(`\n\n\nRebuilding ${JSON.stringify(req.body)} food index\n\n\n`);
```
lukashroch, line 82: Could be remove these additional lines in log msgs, please? Makes the log file non-standard for machine reading.

apps/api/src/jobs/locales/locale-specific-indexes-build.ts
lukashroch, line 1: Could we merge these jobs and make locales optional?

lukashroch (changes requested):

apps/api/.env-template
```diff
@@ -74,6 +74,18 @@ QUEUE_REDIS_HOST=localhost
 QUEUE_REDIS_PORT=6379
 QUEUE_WORKERS=3
 
+# Redis SET
```
lukashroch, line 77: Do we still need the set if it's using cache?

apps/api/src/http/controllers/food-search.controller.ts
```diff
@@ -32,7 +32,23 @@ const ATTR_AS_RECIPE_INGREDIENT_ONLY = 2;
 
 const foodSearchController = ({
   cachedInheritableAttributesService,
-}: Pick<IoC, 'cachedInheritableAttributesService'>) => {
+  subscriberConfig,
+  logger,
+}: Pick<IoC, 'cachedInheritableAttributesService' | 'subscriberConfig' | 'logger'>) => {
+  const redisSubscriber = new RedisSubscriber({ subscriberConfig, logger });
```
lukashroch, line 38: Do we need to create new subscriber here? Imo channel subscription could go to subscriber singleton, init function? Controller will be transient.

apps/api/src/ioc/index.ts
```diff
@@ -251,6 +256,8 @@ export interface IoC extends Jobs {
   mailer: Mailer;
   pusher: Pusher;
   rateLimiter: RateLimiter;
+  redisIndexingSubscriberService: RedisSubscriber;
```
lukashroch, line 259: drop repetitive `redis` ?

apps/api/src/jobs/locales/locale-specific-indexes-build.ts
```diff
@@ -0,0 +1,59 @@
+import type { Job } from 'bullmq';
+
+import type { IoC } from '@intake24/api/ioc';
+
+import BaseJob from '../job';
+
+export default class LocaleSpecificIndexBuild extends BaseJob<'LocaleSpecificIndexBuild'> {
```
lukashroch, line 7: drop the `specific`, something like ` LocaleIndexBuild`

apps/api/src/routes/api/admin/foods/global/index.ts
```diff
@@ -15,5 +15,10 @@ export default () => {
     .get(wrapAsync(adminGlobalFoodsController.read))
     .put(wrapAsync(adminGlobalFoodsController.update));
 
+  // TODO: V4-1016 DELETE ONCE DONE
```
lukashroch, line 18: remove

apps/api/src/routes/api/admin/foods/global/index.ts
```diff
@@ -15,5 +15,10 @@ export default () => {
...
+  // router
+  //   .route('/rebuild')
```
lukashroch, line 20: I would move the API endpoints to `api/admin/locales`, like POST `api/admin/locales/rebuild` as it's locales focused?

lukashroch (changes requested):

apps/api/src/routes/api/admin/foods/global/index.ts
cheburan: This "rebuild all routes" endpoint doesn't make sense anymore since there is no scenario where the admin can trigger a rebuild of all locales. It is still possible via the Specific Locale Rebuild if something will put "all" into the Redis Cache. So we can use this if we would like to give the ability to trigger the rebuild for all locales via the admin dashboard in the future.

apps/api/src/http/controllers/food-search.controller.ts
cheburan: True.

apps/api/src/http/controllers/admin/fdbs/food.controller.ts
```diff
@@ -48,7 +50,10 @@ const adminFoodController = ({
       where: { id: localeId },
     });
 
+    const { code: localeCode } = await resolveLocale(localeId);
+
     const foodLocal = await adminFoodService.createFood(code, req.body);
+    await addToRedisIndexingKeyCache(localeCode, { cache });
```
lukashroch, line 56: Should be keep it within the action that needs the update to happen. It seems it's pulled of out of each action into the  controllers, so will only affect potentially http layer. Though, you probably want to update the index regardless where the update goes through controller/http layer.

apps/api/src/util/redis-indexing.ts
```diff
@@ -0,0 +1,12 @@
+import type { IoC } from '@intake24/api/ioc';
+
+// Add a value to the redisSetService if needed IndexRebuilding
+export const addToRedisIndexingKeyCache = async (value: string, { cache }: Pick<IoC, 'cache'>) => {
```
lukashroch, line 4: Shouldn't this rather be method in cache? Seems kind of an anti-pattern to pull dep out of IoC and run specific thing for cache.

apps/api/src/util/db_resolver.ts
lukashroch, line 1: similarly to `addToRedisIndexingKeyCache`, makes the small util function, dependent on the DB. If this is pulled into the unit testing, it's gonna start failing and will need refactor.

lukashroch (changes requested):

cheburan:
@lukashroch addressed identified issues and refactored the code.

PR #236, opened 2024-02-09
fix(survey): V4-1041: Fixing the overflow and margin left/right issues

cheburan:
Fixing the overflow and margin left/right issue for the v-btn (globally)

### DIFF
--- a/apps/survey/src/components/elements/FoodBrowser.vue
+++ b/apps/survey/src/components/elements/FoodBrowser.vue
@@ -101,7 +101,6 @@
         </v-btn>
         <v-btn
           v-if="type === 'recipeBuilder' && !requiredToFill"
-          class="overflow-button"
           color="primary"
           :disabled="missingDialog"
           large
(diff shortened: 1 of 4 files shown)


apps/survey/src/components/elements/FoodBrowser.vue
lukashroch, line 1: Does it need additional span wrap?

apps/survey/src/components/elements/FoodBrowser.vue
```diff
@@ -109,7 +110,7 @@
           :title="promptI18n['missing.irrelevantIngredient']"
           @click.stop="skipTheStep"
         >
-          {{ promptI18n['missing.irrelevantIngredient'] }}
+          <span>{{ promptI18n['missing.irrelevantIngredient'] }}</span>
```
lukashroch, line 113: Does it need additional span wrap?

lukashroch:
I've made a bit more selective to the btn layout to always select content.

Also removed margin. It's not related to truncation. Also not sure if needed, gaps are done with flex. So this offsets it much more.

PR #246, opened 2024-02-28
feat(admin): Drinkware sets editor

mechkg:
Rework of the drinkware sets editor in the admin tool, adds the following functions:

* Editing and creating new sliding scales in the v2 format
* Converting v1 sliding scales into v2
* Editing and changing the volume measurements


mechkg:
@lukashroch Do you think this is OK to do? https://github.com/MRC-Epid-it24/intake24/blob/admin-drinkware/apps/admin/src/views/images/drinkware-sets/components/SlidingScaleEditor.vue#L305-L310 

It somehow feels like it's dangerous.

lukashroch:
@mechkg Should be fine :-) just keep releasing unused objects with `revokeObjectURL`, I think releasing previous one on change / re-upload and releasing current one on component destroy lifecycle should do the trick - they are tied to `document` object, so they could survive until window reloaded and repetitive uploads will pile up in memory.

PR #259, opened 2024-04-03
fix(api): #V4-464, #V4-1016: stem synonyms

mechkg:


mechkg:
@lukashroch This is a small fix and should be safe to merge

PR #275, opened 2024-05-13
feat(search ranking): add a multiplier to local popularity

mucc001:


apps/api/src/food-index/ranking/local-popularity.ts
```diff
@@ -32,6 +32,7 @@ export async function getLocalPopularityRanking(
 
     rows.forEach((row) => {
       ranking[row.foodCode] = row.occurrences;
+      ranking[row.foodCode] = row.occurrences * row.multiplier;
```
mechkg, line 35: @mucc001 This ranking function caches popularity counter results from the database, you need to keep that in mind when testing (e.g. set stdTTL to 0 here to see changes reflected immediately)

apps/api/src/food-index/ranking/local-popularity.ts
mechkg, line 1: @mucc001 

This ranking function caches counter rows from the database, you need to be mindful of that for testing (e.g. set stdTTL to 0 here to see changes reflected immediately):
https://github.com/MRC-Epid-it24/intake24/blob/cd91e015622ef22fd2583fcfd93a3b720f1e872b/apps/api/src/food-index/ranking/local-popularity.ts#L8

It looks like you're not actually fetching the multiplier column from the database (it's missing from attributes):
https://github.com/MRC-Epid-it24/intake24/blob/cd91e015622ef22fd2583fcfd93a3b720f1e872b/apps/api/src/food-index/ranking/local-popularity.ts#L21-L24

And finally, you also need to cache the modified (i.e., multplied) value in the cache instead of the original counter here:
https://github.com/MRC-Epid-it24/intake24/blob/cd91e015622ef22fd2583fcfd93a3b720f1e872b/apps/api/src/food-index/ranking/local-popularity.ts#L26-L29

mechkg (changes requested):

PR #276, opened 2024-05-14
feat: Locale copy (V4-1166)

lukashroch:


mechkg:
@lukashroch 

On the system database side, food frequency/ranking data (`pairwise_associations_occurrences`, `pairwise_assocations_co_occurrences`, `fixed_food_ranking`) should also be copied, perhaps as an option? People tend to complain about search ordering if this is not initialised.

On the foods database side, there's also `brands` and `food_group_locals`. Not sure what the latter is but I think both should also be copied since they contribute to the local food data.

lukashroch:
@mechkg Thanks for reviewing, I have added all of those now + merged in the PA copy, so should hopefully now be everything.

PR #282, opened 2024-06-06
Clock UI

chrimyn234:
- Implement a Google calendar inspired dropdown style time picker.
- Add new timepickerType field to meal time prompt to select between 'clock' and 'simple' .

### DIFF
--- a/apps/survey/src/components/elements/TimePicker.vue
+++ b/apps/survey/src/components/elements/TimePicker.vue
@@ -1,0 +1,116 @@
+<template>
+  <div>
+    <v-combobox
+      class="cb-time-picker"
+      hide-details
+      :items="timeItems"
+      label="Time"
+      outlined
+      :value="internalTime"
+      @change="updateTime"
+    />
+  </div>
+</template>
+
+<script lang="ts">
+import { computed, defineComponent, ref } from 'vue';
+
+export default defineComponent({
+  props: {
+    value: {
+      type: String,
+      required: true,
+    },
+    allowedMinutes: {
+      type: Number,
+      required: true,
+    },
+    is12HourFormat: {
+      type: Boolean,
+      default: true,
+    },
+  },
+  emits: ['input'],
+  setup(props, { emit }) {
+    const internalTime = ref(formatTime(props.value, props.is12HourFormat));
+    const timeItems = computed(() => generateTimeItems(props.is12HourFormat));
+
+    function generateTimeItems(is12HourFormat: boolean) {
+      const period = is12HourFormat ? ['am', 'pm'] : [''];
+      const hoursInDay = is12HourFormat ? 12 : 24;
+      const minutesInHour = 60 / props.allowedMinutes;
+
+      return period.flatMap(p =>
+        Array.from({ length: hoursInDay }, (_, h) =>
+          Array.from({ length: minutesInHour }, (_, m) => {
+            const hour = is12HourFormat ? (h === 0 ? 12 : h) : h;
+            const formattedHour = String(hour).padStart(2, '0');
+            const formattedMinute = String(m * props.allowedMinutes).padStart(
+              2,
+              '0',
+            );
+
+            return `${formattedHour}:${formattedMinute}${p}`.trim();
+          })).flat(),
+      );
+    }
+
+    function convertTo24HourFormat(time: string) {
+      const [timePart, period] = time.split(/(am|pm)/i);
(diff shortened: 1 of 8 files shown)


packages/common/src/prompts/prompts.ts
```diff
@@ -148,7 +154,15 @@ export type Slider = z.infer<typeof slider>;
...
-  allowedMinutes: z.union([z.literal(1), z.literal(5), z.literal(10), z.literal(15), z.literal(20), z.literal(30)]),
+  allowedMinutes: z.union([
+    z.literal(1),
+    z.literal(5),
+    z.literal(10),
+    z.literal(15),
+    z.literal(20),
+    z.literal(30),
+  ]),
+  timepickerType: z.enum(['clock', 'simple']),
```
lukashroch, line 165: - `timePickerType` is bit repetitive, could we call it e.g. `style`?
- could make the `simple` more descriptive, something like `dropdown` to hit what the style is? ... if we have few more styles in future, simple might quickly collide.

packages/common/src/prompts/prompts.ts
```diff
@@ -207,9 +221,12 @@ const textareaPrompt = baseCustomPrompt.merge(promptValidationProps).extend({
   component: z.literal('textarea-prompt'),
 });
 
-const timePickerPrompt = baseCustomPrompt.merge(promptValidationProps).merge(timePicker).extend({
-  component: z.literal('time-picker-prompt'),
-});
+const timePickerPrompt = baseCustomPrompt
```
lukashroch, line 224: survey frontend timepicker implementation is missing, only `meal-time-prompt` is in PR, can you please add?

apps/survey/src/components/elements/TimePicker.vue
```diff
@@ -0,0 +1,116 @@
+<template>
+  <div>
```
lukashroch, line 2: is the `<div>` wrapper necessary?

apps/survey/src/components/elements/TimePicker.vue
```diff
@@ -0,0 +1,116 @@
...
+    <v-combobox
```
lukashroch, line 3: I think combox also allows to enter values manually / typed-in, if it's intention, you willl need to validate output so the typed-in time format is correct. If only select is intended, simple `<v-select>` can be used

apps/survey/src/components/elements/TimePicker.vue
```diff
@@ -0,0 +1,116 @@
...
+  props: {
+    value: {
+      type: String,
+      required: true,
+    },
+    allowedMinutes: {
+      type: Number,
+      required: true,
+    },
+    is12HourFormat: {
```
lukashroch, line 28: use same format as already defined in prompt props -> `am | pm`, so <component :is> can be used.

apps/survey/src/components/prompts/standard/MealTimePrompt.vue
```diff
@@ -1,8 +1,19 @@
 <template>
-  <card-layout v-bind="{ food, meal, prompt, section, isValid }" @action="action">
+  <card-layout
+    v-bind="{ food, meal, prompt, section, isValid }"
+    @action="action"
+  >
     <v-card-text class="pt-2 meal-time-prompt">
       <v-form @submit.prevent="action('next')">
+        <time-picker
```
lukashroch, line 8: please use dynamic component instead of switch

components can be registered as 
```
components: {clock: 'v-time-picker', simple: 'time-picker'}
```
and then used as 
```
<component :is="prompt.style"  v-model="state" :allowed-minutes="allowedMinutes" :format="prompt.format">
```

packages/common/src/prompts/custom.ts
```diff
@@ -93,6 +93,7 @@ export const timePickerPrompt: Prompts['time-picker-prompt'] = copy({
   type: 'custom',
   id: 'time-picker-prompt',
   name: 'Time picker prompt',
+  timepickerType: 'clock',
```
lukashroch, line 96: can you please add survey frontend impl for the custom time picker?

pnpm-lock.yaml
lukashroch, line 1: I think changes to the lock can be dropped altogether as dependencies are not touched, probably just a rebase artifact

lukashroch (changes requested):
Hi Chris, thanks for the PR!

I've added couple of comments what I would improve / add.

Can you please also add newly added options to the docs (to the admin / prompt types section)? Thanks!

PR #283, opened 2024-06-08
Cli importer csv

cheburan:
Adding the converting tool to the CLI for the food-table CSV conversion to the package format.


cheburan:
[example.zip](https://github.com/user-attachments/files/15756917/example.zip)
Example of the files (CSV and JSON) to use with the cli convertor

apps/api/src/http/controllers/admin/fdbs/category.controller.ts
```diff
@@ -156,6 +157,14 @@ function adminCategoryController({
     res.json(categories);
   };
 
+  const all = async (
```
lukashroch, line 160: This might quickly become a perf hog and cause issues for larger sets, better use paginated query/response?

chrimyn234: Please correct me if I'm wrong, but it looks like this isn't being used anywhere. Also, it looks like browseMain already implements what this is intended to do but with pagination. Should it be removed?

PR #307, opened 2024-09-12
Last-minute fixes and features for the French version

mechkg:
Includes a couple of system-wide features:

- Aggregate choice prompt: final questions about all foods in the recall in a single prompt
- Generalised associated foods prompts: same as the regular associated foods prompt, but with parameters defined at the prompt level rather than the individual food level
- Drink scale v2 prompt now limits the aspect ratio of sliding scales to prevent excessive scrolling (should also be applied to legacy v1 sliding scales, but this is not included in this PR)
- Package import tool now supports guide images (surprisingly wasn't implemented before)

The rest of the changes are minor fixes to the localisation and the French database builder


lukashroch:
Could we add new prompts to the docs?

PR #339, opened 2025-01-31
Dev doc update

anguschiu1:
Made some adjustments on the development docs according to my experience to setup local environment.


apps/docs/config/api/security.md
```diff
@@ -39,11 +39,14 @@ Issuer claim for signing JWT tokens
...
 
-Secret to sign JWT access token
+Secret to sign JWT access token.
+
+(You need to replace the default value, e.g. verybigaccesssecret)
 
 - object-path: `jwt.access.secret`
 - dotenv var: `JWT_ACCESS_SECRET`
 - type: `string`
+- default: `''`
```
lukashroch, line 49: I wouldn't add this. There shouldn't be any default, these secrets should be mandatory, ideally it should throw and not start up at all.

apps/docs/config/api/security.md
```diff
@@ -70,9 +73,12 @@ Lifetime of JWT access token. Defined as `ms-formatted` string, see [ms](https:/
 
 Secret to sign JWT refresh token
 
+(You need to replace the default value, e.g. verybigrefreshsecret)
+
 - object-path: `jwt.admin.refresh.secret`
 - dotenv var: `JWT_ADMIN_REFRESH_SECRET`
 - type: `string`
+- default: `''`
```
lukashroch, line 81: dtto

apps/docs/config/api/security.md
```diff
@@ -161,9 +167,12 @@ Lifetime of JWT access token. Defined as `ms-formatted` string, see [ms](https:/
 
 Secret to sign JWT refresh token
 
+(You need to replace the default value, e.g. verybigrefreshsecret)
+
 - object-path: `jwt.survey.refresh.secret`
 - dotenv var: `JWT_SURVEY_REFRESH_SECRET`
 - type: `string`
+- default: `''`
```
lukashroch, line 175: dtto

apps/docs/config/api/security.md
anguschiu1: Acknowledged, I will revert all these lines, and convey the meaning of "JWT secrets have to be provided specifically." as comments in the `.env-template`, so that newcomers will know they have to fill in before starting the API server.

lukashroch: There are actually two cli commands, so we could also include links to it:
- https://docs.intake24.org/cli/generate-env
- https://docs.intake24.org/cli/generate-key

- `generate-env` creates `.env` files from templates and generates the secret on the fly
- whole `generate-key` is just a helper to get a key in console out

anguschiu1: I will add back these links to the docs.

However, the commands `generate-env` and `generate-key` run under `/intake24/apps/cli` throws with errors: 
```
❯ pnpm cli generate-key

> intake24@2024.8 cli /Users/anguschiu/dev/intake24
> pnpm --filter @intake24/cli cli "generate-key"


> @intake24/cli@2024.8 cli /Users/anguschiu/dev/intake24/apps/cli
> node dist/cli.js "generate-key"

Error: option creation failed due to '-sf' in option flags '-sf, --skip-foods [food-ids...]'
- a short flag is a single dash and a single character
  - either use a single dash and a single character (for a short flag)
  - or use a double dash for a long option (and can have two, like '--ws, --workspace')
    at splitOptionFlags (/Users/anguschiu/dev/intake24/node_modules/.pnpm/commander@13.1.0/node_modules/commander/lib/option.js:342:13)
    at new Option (/Users/anguschiu/dev/intake24/node_modules/.pnpm/commander@13.1.0/node_modules/commander/lib/option.js:20:25)
    at run (/Users/anguschiu/dev/intake24/apps/cli/dist/cli.js:62629:29)
    at 49265 (/Users/anguschiu/dev/intake24/apps/cli/dist/cli.js:62734:1)
    at __webpack_require__ (/Users/anguschiu/dev/intake24/apps/cli/dist/cli.js:147559:42)
    at /Users/anguschiu/dev/intake24/apps/cli/dist/cli.js:147611:37
    at Object.<anonymous> (/Users/anguschiu/dev/intake24/apps/cli/dist/cli.js:147613:12)
    at Module._compile (node:internal/modules/cjs/loader:1565:14)
    at Object..js (node:internal/modules/cjs/loader:1708:10)
    at Module.load (node:internal/modules/cjs/loader:1318:32)
/Users/anguschiu/dev/intake24/apps/cli:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @intake24/cli@2024.8 cli: `node dist/cli.js "generate-key"`
Exit status 1
```
It seems the problem is on `cli.js` but I cannot find the root cause. Can you reproduce it at your end?

lukashroch: I've fixed that now. It was due to colliding short flags, which now throws error after commander lib update.

anguschiu1: Code updated, link to CLI page added.

apps/docs/config/api/security.md
anguschiu1: Code updated, link to CLI page added.

apps/docs/config/api/security.md
anguschiu1: Code updated, link to CLI page added.

lukashroch (approved):

PR #343, opened 2025-02-05
test: Add GTM and MS Clarify injection code

anguschiu1:
Please help to deploy to [DEV](https://it24-dev-admin.mrc-epid.cam.ac.uk/) env, so that I can further configure and test with GTM and MS Clarity.


lukashroch:
@anguschiu1 gtag is already included and hooked into the both apps (https://github.com/MRC-Epid-it24/intake24/blob/main/apps/survey/src/main.ts#L31). It's using this plugin (https://matteo-gabriele.gitbook.io/vue-gtag) and DEV/LIVE should be already connected to property tags in GA Analytics console.

Does clarity.ms need and extra script setup in code? I think it can be given access to GA property in console and it fetches the data from there? I know you can do both, but just wondering if it's worth polluting the code with this as first approach might be cleaner? If not I think we'll need to integrate it into vue framework with also cookie consent.

anguschiu1:
The clarity.ms allows visualized session recordings and replays for free, which is not offered by GA and should be quite useful. They also provided integration strategies with framework and cookie consent.

As I just find a way to test GTM and Clarity locally, I have no urgent to use DEV to test it out, I will close this request now and make another pull request later.

anguschiu1:
@lukashroch indeed you're correct, MS clarity can fetch data from GA and hence we need not to setup scripts for that. (I cannot make it work it until today, as platforms require time to sync up).
Now we only need to integrate Google Tag Manager with consent, and it's set. GTM still have a role here as it offers more features to both dev (e.g. preview mode allows testing tags locally) and tag management. 

I will submit pull request for GTM integration to Vue framework with cookie consent.

apps/survey/.env-template
```diff
@@ -40,6 +40,9 @@ VITE_CAPTCHA_SITEKEY=
 VITE_GOOGLE_ANALYTICS_ID=
```
lukashroch, line 40: Can you put GA and GTM IDs close together so both tracking IDs are close, will be easier to navigate.

apps/survey/src/main.ts
```diff
@@ -31,6 +31,7 @@ app.use(vuetify);
 app.use(VueGtag, { bootstrap: false }, router);
 app.use(cookieConsentPlugin, cookieConsentConfig());
 app.use(createManager({ deferLoading: { enabled: true, autoLoad: true } }));
+app.use(createGtm({ id: import.meta.env.VITE_GTM_CONTAINER_ID, enabled: false, debug: true, vueRouter: router }));
```
lukashroch, line 34: I'd enable debug only in DEV mode, so it doesn't get dragged into production.

```ts
debug: import.meta.env.DEV
```

packages/ui/src/cookie-consent/config.ts
```diff
@@ -21,6 +22,16 @@ async function toggleGA(enabled: boolean) {
   setOptions(gTagConfig());
   await bootstrap();
 }
+async function toggleGTM(enabled: boolean) {
+  const gtm = useGtm();
+
+  if (!enabled) {
+    gtm?.enable(false);
```
lukashroch, line 29: Probably not needed condition statement?
```ts
gtm?.enable(enabled);
```

lukashroch (changes requested):
I've added couple of tweaks if OK?

Can you please push the branch into intake24 remote instead of your personal remote? It will be easier to rebase to `dev`

apps/survey/src/main.ts
anguschiu1: It's very helpful, thank you!

packages/ui/src/cookie-consent/config.ts
anguschiu1: You're right. I'll revise it.

anguschiu1:
> I've added couple of tweaks if OK?
> 
> Can you please push the branch into intake24 remote instead of your personal remote? It will be easier to rebase to `dev`

anguschiu1:
> > I've added couple of tweaks if OK?
> > Can you please push the branch into intake24 remote instead of your personal remote? It will be easier to rebase to `dev`

Sure. Let me open another pull request from intake24 remote and accommodate the above change request.

PR #352, opened 2025-02-24
Integrate GTM and MS Clarity to support user behaviour analysis (V4-1409)

anguschiu1:
The change involved in these areas:
1. Install, load and initialize `@gtm-support/vue-gtm` to web app with corresponding GTM container ID.
3. Install, load and initialize `@microsoft/clarity` to web app with corresponding MS Clarity project ID.
4. Implement to seek cookie consent from users when using GTM and MS Clarity.
5. Implement GTM custom events in several prompts.
6. Implement search counts and scroll percentage for custom events.


apps/survey/src/components/handlers/standard/MealTimePromptHandler.vue
```diff
@@ -52,6 +53,15 @@ export default defineComponent({
       if (type === 'cancel') {
         survey.deleteMeal(meal.value.id);
         ctx.emit('action', 'next');
+
```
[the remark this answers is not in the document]
mechkg: If I am not mistaken, the requirement from the dietary team was to track meal deletion via "I did not have X" button in the meal time prompt vs the delete meal button in the quick list/review screen, so there should be a way to differentiate between those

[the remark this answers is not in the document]
lukashroch: Yeah, `cancel` could still be passed up to the recall though (with the additional context)? It's probably just worth refactoring as it snowballs things like this, forcing to capture GTM here explicitly rather than in one place in top-level recall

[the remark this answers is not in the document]
anguschiu1: The current requirement is to know how people delete a meal, whether they click "did not have" at the meal-time prompt or "delete meal" after the meal is set (edit-meal prompt). 

Would you enlighten me if I can get the details like meal name and the click action, and fire GTMs at the closer place?

apps/survey/src/components/recall/recall-mixin.ts
```diff
@@ -295,6 +301,15 @@ export default defineComponent({
           break;
         case 'deleteMeal':
           this.survey.deleteMeal(mealId);
+
```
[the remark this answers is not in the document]
anguschiu1: Yes, it's probably going to extend and need to refit here and there, once the team start to know how to look into GA reports. As I still not sure which data will need to capture into custom events, my idea is to standardise the event generation, while do minimal change on the code base.

For now GTM has built-in function to trigger and fire [history-change](https://support.google.com/tagmanager/answer/7679322?hl=en) events for SPA. I have set it up and it will be shown as "page-view" events in GA, with `prompt_id` (e.g. "instructions-carousel") and `previous_prompt_id` as event parameters.

image

This approach should cover most page change with general information available for GA reporting.

However, for more refined information like scroll percentages, search terms and search counts, we may still need to implement code for custom events.

But I agreed that we should type the GTM events and use your convention for `action` and `events`

apps/survey/src/components/handlers/standard/MealTimePromptHandler.vue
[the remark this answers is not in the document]
lukashroch: It depends were we draw the line between presentation logic and business log. I guess from my point of view top-level / start of presentation logic is the recall components (`recall-mixin.ts`), which drills down to handlers -> prompt -> elements. While business logic is the `dynamic-recall (sry/dynamic-recall)` and attached stores. So I would put as much logic to top (`action(...)` | `mealAction(...)` | `foodAction(...)` in `recall-mixin.ts`), which should catch most of the actions. And what cannot be caught there, then drill down to components?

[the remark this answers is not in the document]
lukashroch: @anguschiu1 just realized that top-level approach could actually swallow more use-cases than I thought as quite few actions prompt actions are intercepted and processed in handler and only `next` is send up to top-level. So maybe better approach would be log the event in `usePromptUtils` - this composable is used for most of the prompt and so loggin GTM here (https://github.com/MRC-Epid-it24/intake24/blob/main/apps/survey/src/composables/use-prompt-utils.ts#L107) might be better option?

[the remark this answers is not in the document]
anguschiu1: My bad, I see this message just now, and I have implemented on `recall-mixin.ts` for testing already (and found mostly `next` action, like you said)

I will learn how to move them into`use-prompt-utils.ts`, but for the events that expects scroll percentage and search terms counts, perhaps I should keep it on presentation logic, or to push these details upwards?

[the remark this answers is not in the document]
lukashroch: > I will learn how to move them into `use-prompt-utils.ts`

It should just be moving the event firing from recall-mixin/handlers to the `action` in the composable  `use-prompt-utils.ts` - that should capture lot of actions in prompts.

> but for the events that expects scroll percentage and search terms counts, perhaps I should keep it on presentation logic, or to push these details upwards?

Yep, that's fine to keep where it is now

apps/survey/src/composables/use-prompt-utils.ts
```diff
@@ -106,6 +108,78 @@ export function usePromptUtils<
...
+          gtmEventParams.event = GtmEvents.SelectFoodCategory;
+          break;
+        case 'restart':
+          gtmEventParams.action = type;
+          gtmEventParams.event = GtmEvents.Restart;
+          break;
+      }
+      if (gtmEventParams.event) {
+        sendGtmEvent(gtmEventParams);
+      }
```
anguschiu1, line 182: I cannot trigger all actions to verify all event triggering, although I think it will still work anyway.
If there are any neat way to minimise the switch statement, please advise!

packages/ui/src/tracking/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  label?: string; // auxillary data for custom events
+  search_term?: string;
+  search_count?: number;
+  percent_scrolled?: number; // 0-100 percent scrolled refers to the percentage of the page that has been viewed so far
+  noninteraction: boolean; // false if an event is not tracked as a user-generated action
+};
+
+export function sendGtmEvent(params: GtmEventParams): void {
+  useGtm()?.trackEvent(params);
+}
```
anguschiu1, line 51: I used object literals to replace enums.

packages/ui/src/tracking/index.ts
```diff
@@ -0,0 +1,51 @@
+import { useGtm } from '@gtm-support/vue-gtm';
+
+export const GtmEvents = {
```
lukashroch, line 3: use array, most of is just key:value object duplication of actionTypes so you can expand on that:
```
export const gtmEvents = [...actionType, 'one more', 'more', 'etc'], as const;
export type GtmEvent = (typeof gtmEvents)[number];
```
then you can simplify the switch as it can fallthrough if you're only collecting type.

packages/ui/src/tracking/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  SelectFood: 'selectFood',
+  SelectMeal: 'selectMeal',
+  SelectFoodCategory: 'selectFoodCategory', // it is a action not found in action type
+  Next: 'next',
+  Restart: 'restart',
+} as const;
+export type GtmEvent = typeof GtmEvents[keyof typeof GtmEvents];
+
+// Event category for Google Analytics, deduced from Scheme prompts structure
+export const GtmSchemePrompts = {
```
lukashroch, line 23: same as above, use array, object of same key:value is superfluous

apps/survey/src/components/elements/FoodBrowser.vue
```diff
@@ -223,6 +225,25 @@ const recipeBuilderFoods = ref<FoodHeader[]>([]);
 const recipeFoods = ref<RecipeFood[]>([]);
 const recipeBuilderToggle = ref(false);
 const tab = ref(0);
+const searchCount = ref(1);
+const percentScrolled = ref(0);
+
+function onScroll(event: Event) {
```
lukashroch, line 231: no sure how often this fires up (would need to check the directive it has has option to throttle) but if a lot, it might be worth to throttle (vueUse lib probably has utility for that), but can wait...

lukashroch (changes requested):

apps/survey/src/components/elements/FoodBrowser.vue
anguschiu1: It fires continuously when scrolling, but the function will only update the reactive `percentScrolled` on certain percentile.
I just checked that there is a [usethrottlefn](https://vueuse.org/shared/useThrottleFn/#usethrottlefn), we can use it if any issues arise around this.

apps/survey/src/composables/use-prompt-utils.ts
anguschiu1: Revised, but for the `type==='cancel'` scenario, I don't want to access `GtmEvents[key]` using `key` directly, so use a loop for that. (If there's a better way please let me know)

lukashroch: you can simplify this:

from
```
for (const key in GtmEvents) {
        if (type === GtmEvents[key]) {
          gtmEventParams.event = GtmEvents[key];
        }
        if (type === 'cancel') { // Cancel button
          for (const key in GtmEvents) {
            if (GtmEvents[key] === 'deleteMeal') {
              gtmEventParams.event = GtmEvents[key];
            }
          }
        }
      }
```

to
```
// you know it's valid based on this check already, no need to loop
if (type === 'cancel') {
  gtmEventParams.event = 'deleteMeal'
}
gtmEventParams.event = GtmEvents.includes(type) ? type : undefined

```

apps/survey/src/components/elements/CategoryContentsView.vue
```diff
@@ -129,10 +138,46 @@ export default defineComponent({
...
     categorySelected(category: CategoryHeader): void {
       this.$emit('category-selected', category);
+      const params: GtmEventParams = {
+        food_category: category.name,
+        search_term: this.searchTerm,
+        search_count: this.searchCount,
+        percent_scrolled: this.percentScrolled,
+        noninteraction: false,
+      };
+      for (const key in GtmEvents) {
```
lukashroch, line 148: this looping is unnecessary, types you establish already guard that so you can just assign it:
```
params.event = 'selectFoodCategory;
....
params.scheme_prompts = 'foods'
```

apps/survey/src/components/elements/CategoryContentsView.vue
```diff
@@ -129,10 +138,46 @@ export default defineComponent({
...
+        if (GtmEvents[key] === 'selectFoodCategory') {
+          params.event = GtmEvents[key];
+        }
+      }
+      for (const key in GtmSchemePrompts) {
```
lukashroch, line 153: dtto

apps/survey/src/components/elements/CategoryContentsView.vue
```diff
@@ -129,10 +138,46 @@ export default defineComponent({
...
     foodSelected(food: FoodHeader): void {
       this.$emit('food-selected', food);
+      const params: GtmEventParams = {
+        food: food.name,
+        search_term: this.searchTerm,
+        search_count: this.searchCount,
+        percent_scrolled: this.percentScrolled,
+        noninteraction: false,
+      };
+      for (const key in GtmEvents) {
```
lukashroch, line 170: dtto

apps/survey/src/components/elements/CategoryContentsView.vue
```diff
@@ -129,10 +138,46 @@ export default defineComponent({
...
+        if (GtmEvents[key] === 'selectFood') {
+          params.event = GtmEvents[key];
+        }
+      }
+      for (const key in GtmSchemePrompts) {
```
lukashroch, line 175: dtto

lukashroch (changes requested):

lukashroch (changes requested):

apps/survey/src/components/elements/CategoryContentsView.vue
```diff
@@ -129,10 +138,30 @@ export default defineComponent({
   methods: {
     categorySelected(category: CategoryHeader): void {
       this.$emit('category-selected', category);
+      const params: GtmEventParams = {
+        event: gtmEvents.includes('selectFoodCategory') ? 'selectFoodCategory' : '',
```
lukashroch, line 142: Does these conditions make sense? I mean `gtmEvents` and `gtmSchemePrompts` are constants, you know it's always there. And we don't want to send empty event anyone?

So should be 
```
event: 'selectFoodCategory'
```

apps/survey/src/components/elements/CategoryContentsView.vue
```diff
@@ -129,10 +138,30 @@ export default defineComponent({
...
+        scheme_prompts: gtmSchemePrompts.includes('foods') ? 'foods' : '',
```
lukashroch, line 143: dtto
```
scheme_prompts: 'foods' ,
```

apps/survey/src/components/elements/CategoryContentsView.vue
```diff
@@ -129,10 +138,30 @@ export default defineComponent({
...
+        percent_scrolled: this.percentScrolled,
+        noninteraction: false,
+      };
+      sendGtmEvent(params);
     },
 
     foodSelected(food: FoodHeader): void {
       this.$emit('food-selected', food);
+      const params: GtmEventParams = {
+        event: gtmEvents.includes('selectFood') ? 'selectFood' : '',
```
lukashroch, line 156: dtto
```
event: 'selectFood' ,
```

apps/survey/src/components/elements/CategoryContentsView.vue
```diff
@@ -129,10 +138,30 @@ export default defineComponent({
...
+        scheme_prompts: gtmSchemePrompts.includes('foods') ? 'foods' : '',
```
lukashroch, line 157: dtto
```
scheme_prompts: 'foods'
```

lukashroch (changes requested):

apps/survey/src/components/elements/CategoryContentsView.vue
anguschiu1: I thought we prefer to use this syntax to type this `GtmEventParams.event`, but empty events is confusing, just denoting something is off.
But then, there will no situation we will use the type `GtmEvent` for `GtmEventParams.event`. Should I keep it, or remove it?

lukashroch: Definitely keep the types, it's part of:
```
export type GtmEventParams = {
  event?: GtmEvent | string;
  ...
}
```

It's just loosely typed with `GtmEvent | string;` because those actions in code are also loosely typed (`type: string`) at the moment. Ideally we should improve action types so they are more strictly typed in future, then we'll be able to remove this loose type too.

lukashroch (approved):

PR #360, opened 2025-04-03
Gousto locale features

mechkg:
The main added feature is the support for food thumbnail images, but this PR also includes multiple fixes for issues found during tested, mostly related to search restricted to a category.


apps/survey/src/components/elements/YesNoChoice.vue
```diff
@@ -0,0 +1,60 @@
...
+      <v-icon :color="modelValue === true ? 'primary' : 'secondary '" start>
+        $yes
+      </v-icon>
+      {{ $t('common.action.yes') }}
+    </v-btn>
+  </v-btn-toggle>
+</template>
+
+<script lang="ts">
+/*
```
lukashroch, line 23: IMO this is an issue down to props.mandatory logic and how it's passed down to toggle component, which might not be correct so would need proper fix?

... anyway, this just seems to drop the mandatory logic (although prop is still there and not used) and emit directly, shall fix or delete the original so we don't have a dead code?

apps/api/src/http/routers/admin/food-thumbnail-images.router.ts
```diff
@@ -0,0 +1,48 @@
...
+          user,
+          foodThumbnailImageService,
+          i18nService,
+        } = req.scope.cradle;
+
+        const res = imageMulterFile.safeParse(file);
+
+        if (!res.success) {
+          // i18n helper functions expect Express req while req here is TsRestRequest
+          throw new ValidationError(
```
lukashroch, line 32: I tried to rework this recently so we don't have fiddle to much with i18n in controllers.
You can do now this 
```
throw ValidationError.from({ path: 'image', i18n: { type: 'file._' } });
```
and i18n stuff gets resolved at the end in error handler

apps/survey/src/components/prompts/standard/FoodSearchPrompt.vue
```diff
@@ -45,14 +45,17 @@ const { action } = usePromptUtils(props, { emit });
 
 const isValid = true;
 const rootCategory = computed(() => {
-  const foodSearch = props.meal?.flags?.find(flag => flag.startsWith('food-search:'))?.split(':')[1];
+  const mealFlagCategory = props.meal?.flags?.find(flag => flag.startsWith('food-search:'))?.split(':')[1];
```
lukashroch, line 48: might be worth to use same naming convention (whichever one) for consistency? When being set up later, convention is easier to follow.

apps/survey/src/components/elements/YesNoChoice.vue
mechkg: I think the mandatory flag has slightly different behaviour here, if I understand it correctly. The default behaviour of `v-btn-toggle` allows undefined state, and a choice can be undone (reset back to undefined) by clicking on the currently selected button again. 

The mandatory flag makes it so that there is no undefined state, and one of the buttons is pre-selected and cannot be unselected.

The Yes/No prompt logic ignores switching to undefined state but automatically proceeds to the next prompt if the value is defined, and if it starts with "No" pre-selected (by going back to a previously answered prompt, for example) clicking on "No" again to confirm it will unselect it instead.

What I wanted here is for this component to start in the undefined state, but once an answer is selected it can no longer be reset to undefined because it doesn't make a lot of sense for this type of prompt.

apps/survey/src/components/prompts/standard/FoodSearchPrompt.vue
mechkg: You're right, I didn't want to break any existing schemes by renaming the food flag and the new category flag has already been used as well so it ended up like this :\

apps/api/src/http/routers/admin/food-thumbnail-images.router.ts
mechkg: Thanks! Changed to use this new function :)

PR #389, opened 2025-07-30
feat(gtm): Added uxSessionId, uxUserId, search terms to GTM

anguschiu1:
The change is to include `uxSessionId` and `userId` (renamed as `uxUserId` in GTM as it’s reserved by GA) in every GTM custom event. This allows for better correlation of data collected in GA with data in Intake24 databases.

Additionally, a new `foodSearch` GTM custom event has been added. This event fires every `search()` query triggered in `foodBrowswer`, capturing any food text search that triggers a food catalog rebuild and the number of foods displayed in each catalog.

Lastly, a minor fix to add a reset for `percentScrolled` counter when `search()` triggered.


packages/ui/src/tracking/index.ts
```diff
@@ -34,6 +34,10 @@ export type GtmEventParams = {
   search_count?: number;
   percent_scrolled?: number; // 0-100 percent scrolled refers to the percentage of the page that has been viewed so far
   noninteraction: boolean; // false if an event is not tracked as a user-generated action
+  uxSessionId?: string | ''; // unique session id generated by the survey app to help correlating DB records
```
lukashroch, line 37: do we need empty string type here? I think `uxSessionId?: string` should be enough?

packages/ui/src/tracking/index.ts
```diff
@@ -34,6 +34,10 @@ export type GtmEventParams = {
...
+  uxUserId?: string | ''; // user id generated by in the backend
```
lukashroch, line 38: dtto

lukashroch (changes requested):
Hi Angus,

few thoughts

- do we need to inject the userId? We should be able to link with user through the sessionId I think. Though in general if we are after not-submitted recall, i.e. partial recalls, we should probably re-visit sessions implementation. It wasn't much improved since v3, and at the moment user can only have one and is overridden, so partial recall potentially lost, which can probably be problematic to get hold of the session and then even the user record.

- are these really globally used for all gtm events? From first glance looks like we are injecting into to the foods browser and then emitting to gtm. I'm thinking if it's possible to add some global metadata to each gtm (like into the vue-gtag). If not, I think we could wrap the vue-gtm with some simple our own wrapper that would handle attaching all the global metadata, like these sessionId etc.

packages/ui/src/tracking/index.ts
anguschiu1: I revised it already.

anguschiu1:
Hi @lukashroch ,

> * do we need to inject the userId? We should be able to link with user through the sessionId I think. Though in general if we are after not-submitted recall, i.e. partial recalls, we should probably re-visit sessions implementation. It wasn't much improved since v3, and at the moment user can only have one and is overridden, so partial recall potentially lost, which can probably be problematic to get hold of the session and then even the user record.

I think there are two benefits to inject `userId` together with `uxSessionId`:
1. DA team can create reports in GA with the dimension of `uxUserID` to do some quick check, without the need to exporting and cross-checking with Intake24 database
2. I think it should be able to make uxUserId as a filtering criteria in MS Clarity, DA team members can search for screen recording a little easier.

> * are these really globally used for all gtm events? From first glance looks like we are injecting into to the foods browser and then emitting to gtm. I'm thinking if it's possible to add some global metadata to each gtm (like into the vue-gtag). If not, I think we could wrap the vue-gtm with some simple our own wrapper that would handle attaching all the global metadata, like these sessionId etc.

Agreed, I added a `gtmEvent.ts` as util, and refactored some codes s.t. we can consolidate global metadata such as uxSessionId and userId in there.

apps/survey/src/components/elements/CategoryContentsView.vue
```diff
@@ -181,6 +181,14 @@ export default defineComponent({
       type: Number as PropType<number>,
       default: 0,
     },
+    uxSessionId: {
```
lukashroch, line 184: redundant, remove

apps/survey/src/components/elements/CategoryContentsView.vue
```diff
@@ -181,6 +181,14 @@ export default defineComponent({
...
+      type: String as PropType<string>,
+      required: true,
+    },
+    userId: {
```
lukashroch, line 188: redundant, remove

apps/survey/src/components/elements/FoodBrowser.vue
```diff
@@ -80,6 +80,8 @@
           :grid-threshold="prompt.gridThreshold"
           :i18n="promptI18n"
           :type="type"
+          :user-id="userId"
```
lukashroch, line 83: redundant, remove

apps/survey/src/components/elements/FoodBrowser.vue
```diff
@@ -99,6 +101,8 @@
           :search-count="searchCount"
           :search-term="searchTerm ?? undefined"
           :type="type"
+          :user-id="userId"
```
lukashroch, line 104: redundant, remove

apps/survey/src/components/elements/FoodBrowser.vue
```diff
@@ -233,7 +239,9 @@ const props = defineProps({
 });
 
 const emit = defineEmits(['action', 'foodSelected', 'foodMissing', 'recipeBuilder', 'update:modelValue', 'foodSkipped']);
-
+const survey = useSurvey();
+const uxSessionId = ref('');
```
lukashroch, line 243: redundant, remove

apps/survey/src/components/elements/FoodBrowser.vue
```diff
@@ -233,7 +239,9 @@ const props = defineProps({
...
+const userId = ref('');
```
lukashroch, line 244: redundant, remove

apps/survey/src/util/gtmEvent.ts
```diff
@@ -0,0 +1,22 @@
+import { useGtm } from '@gtm-support/vue-gtm';
+import { useSurvey } from '@intake24/survey/stores';
+import type { GtmEventParams } from '@intake24/ui/tracking';
+
+export function sendGtmEvent(params: GtmEventParams): void {
+  const eventParams = { ...params };
+  eventParams.noninteraction = false;
+  try {
+    const survey = useSurvey();
+    if (survey) {
```
lukashroch, line 10: if check should not be needed, I'd refactor to something like this to be clear what overrides what.

```ts
try {
    const survey = useSurvey();

    const defaults = {
      noninteraction: false,
      uxSessionId: survey.data.uxSessionId,
      uxUserId: survey.user?.userId,
    };

    useGtm()?.trackEvent({ ...defaults, ...params });
  }
  catch (e) {
    console.error('Error sending GTM event:', e);
  }
```

lukashroch (changes requested):
@anguschiu1 OK, sounds good, let's send both then.

I think we can clean this up a bit since it's handled in wrapper, please see few comments.

PR #403, opened 2025-09-23
Drop prototype feature

mechkg:
@lukashroch Please see the suggested changes to this migration:

- Split global/local merge and removal of the prototype feature into separate migrations
- Reorder operations to do the prototype resolution/removal first (easier to do it on the old db structure)
- Re-implement prototype resolution to (hopefully) match the intended behaviour


mechkg:
@lukashroch My bad, I had removed the part that had to do with the inheritance from the original migration but somehow didn't commit it, pushed it now.

PR #412, opened 2025-11-08
Clean up circular imports

mechkg:
1. Added circular-dependency-plugin for Webpack

Plugin is configured to report circular imports as warnings during Webpack build process (only in dev configuration)

Added a custom (fairly hacky) filter for circular-dependency-plugin to identify types used in Sequelize models decorators (`@HasMany`, `@BelongsTo` etc.) and suppress circular import warnings because this is safe

2. Cleaned up circular imports in Sequelize models: changed imports from barrel files to direct local imports
3. Resolved some circular dependencies between IoC/service definition files

### DIFF
--- a/apps/api/circular-dependency-utils.js
+++ b/apps/api/circular-dependency-utils.js
@@ -1,0 +1,123 @@
+/* eslint-disable regexp/no-trivially-nested-quantifier */
+/* eslint-disable regexp/no-super-linear-backtracking */
+/* eslint-disable style/no-multi-spaces */
+
+import fs  from 'node:fs';
+import path from 'node:path';
+
+const sequelizeDecorators = [
+  'HasMany',
+  'BelongsTo',
+  'HasOne',
+  'BelongsToMany',
+  'Scope',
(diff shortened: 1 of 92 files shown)


mechkg:
Hi @lukashroch, 

The changes to Sequelize model files are straightforward and are mostly to tidy up cyclic imports involving barrel files (afaik none of them were causing any actual problems but as far I understand this is still a bad practice and potentially dangerous). I dropped the idea of factoring the associations decorators out to separate files to avoid circular imports in Sequelize models and this is now handled by a custom filter in circular-dependency-plugin config.

Please have a look at these changes:

https://github.com/intake24/intake24/commit/c12827bbed23a5e4fb84ac1b1c7d9f118b72e015

I think passing the process() function to Workers without binding this was incorrect because they were accessing this.name, could you confirm that?

I have changed the dependency on runtime DI resolution (by job name) so it is explicitly passed to the constructor instead of  relying on the global import of the DI container. I don't think it caused any runtime problems because it was accessed late but it was an actual circular dependency.

PR #482, opened 2026-05-06
fix(admin): V4-1711: implement reset function when clearing feedback...

serjoshua:
fix(admin): V4-1711: implement reset function when clearing feedback scheme field

### DIFF
--- a/apps/admin/src/views/surveys/form.vue
+++ b/apps/admin/src/views/surveys/form.vue
@@ -581,6 +581,7 @@
                 :label="$t('feedback-schemes._')"
                 name="feedbackSchemeId"
                 resource="feedback-schemes"
+                @click:clear="resetFeedback"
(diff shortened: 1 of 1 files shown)


apps/admin/src/views/surveys/form.vue
```diff
@@ -581,6 +581,7 @@
                 :label="$t('feedback-schemes._')"
                 name="feedbackSchemeId"
                 resource="feedback-schemes"
+                @click:clear="resetFeedback"
```
lukashroch: I'd use the ` @update:model-value` event. `@click` can be disjointed from the actual updated value (i.e. don't know what the value is), while in `@update:model-value` you have access to value and only reset it when it's false.

lukashroch (changes requested):

lukashroch (approved):

PR #490, opened 2026-06-01
fix(admin): [V4-1001] create CSS classes for food list components

serjoshua:


serjoshua:
@lukashroch I pushed some changes from last week that has been tested already in Dev. Let me know if the changes are ok.

apps/admin/src/app.vue
```diff
@@ -226,4 +226,8 @@ onMounted(async () => {
 
 <style lang="scss">
 @use '@intake24/admin/scss/app.scss';
+
+a.v-list-item {
```
lukashroch: I think this is not needed

packages/ui/src/components/layouts/app-footer.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <v-divider v-if="$vuetify.display.mobile" />
-  <v-footer v-if="!$vuetify.display.mobile || !loggedIn" class="justify-center pa-4 flex-grow-0">
+  <v-footer v-if="!$vuetify.display.mobile || !loggedIn" id="footer" app class="justify-center pa-4 flex-grow-0" style="position: absolute;">
```
lukashroch: is the `app` and `style` needed? If we need style I think there is an utility class for that

apps/admin/src/views/fdbs/read.vue
```diff
@@ -1,11 +1,11 @@
 <template>
   <entry-layout v-if="entryLoaded" v-bind="{ id, entry }">
-    <v-row class="justify-space-between pa-4">
-      <v-col cols="5" lg="4" xl="3">
+    <v-row class="justify-space-between py-4">
+      <v-col cols="5" lg="4" xl="4">
```
lukashroch: `xl="4"` is redundant here

lukashroch (commented):
lgtm, just few not used leftovers to remove

apps/admin/src/views/fdbs/read.vue
serjoshua: So for XL screens, the total column number previously is 11 (8 + 3) so it used to look like this:

Image

For XL screens after this change, it looks like this now:

Image

Large screens have no issue because total columns is 12 already. I just thought that the gap in XL screens didn't look great. I can change it back if needed.

apps/admin/src/app.vue
serjoshua: This is tricky. This is connected to menu-item being _scoped_ because Account menu in the side bar is a `v-list`, while the rest of the side bar is a `menu-tree`. I have to put this style to keep the Account items not being indented that far. Otherwise, it messes up the food list if I don't put menu-item to be _scoped_.

Image

packages/ui/src/components/layouts/app-footer.vue
serjoshua: This is a really good question. Our footer previously looks like this in small screens.

Image

The side bar looks like it is eating up the footer. If I put `app`, the footer becomes a part of the main screen and centres the footer view properly. The problem is, it becomes a sticky footer. It doesn't go away.

Image

So, I have to make it absolute so it will behave like before but is now centred.

Image

These are changes that indirectly affects the scrolling and the food list view that is why I thought of fixing them as well. Also, Suzanna made some comments with the indentation last time,

lukashroch: OK, though I would use the class instead of style anyway if there is a utility class for that. 

... I think we should get rid of the footer and only display it on screens when user is not authenticated in order to get more space in the screens anyway.

apps/admin/src/app.vue
lukashroch: The excessive padding is handled by scoped scss in menu-item, which seems to be working for me, having this on top doesn't do anything for me.

apps/admin/src/views/fdbs/read.vue
lukashroch: The change is fine, but since both values are the same now, i.e. `lg="4" xl="4"`, you can only specify the smaller screen `lg="4"`, it has the same affect.

lukashroch (changes requested):

apps/admin/src/app.vue
serjoshua: I can deploy it to Dev so you could try. I tried deploying it to Dev many times and it ruins the Account menu for me. What do you think?

packages/ui/src/components/layouts/app-footer.vue
serjoshua: There might be a position attribute. I will do that. As for removing the footer, would you like to include it in this change?

apps/admin/src/views/fdbs/read.vue
serjoshua: I see. I will test this and remove xl then.

serjoshua (commented):
Refactor changes in footer and in side bar menu

apps/admin/src/app.vue
```diff
@@ -170,6 +148,11 @@ useLanguage('admin', http, vI18n);
 const sidebar = ref(display.lgAndUp);
 const resources = groupBy(defaultResources, 'group');
 
+const userDashboardMenuItem = { name: 'dashboard', icon: 'fas fa-tachometer-alt', title: 'dashboard._' } as any;
```
lukashroch: Let's not do `any` if possible :-)

I think you can inline these into the template and you can loosen the `item` type in `menu-item` component.

At the moment it has `Resource` type, so this can be loosed into something like
```ts
type MenuItem = {
  name:string;
  icon: string;
  title: string;
}
```

lukashroch (changes requested):

apps/admin/src/app.vue
lukashroch: just change item prop type to something like
```ts
type MenuItem = {
  name:string;
  icon: string;
  title?: string;
}
```
the item menu-item doesn't need full `Resource` type

lukashroch (approved):

PR #498, opened 2026-06-22
feat(survey): update food based on prompt answers

miles-eastwood:
- Adds the ability to swap out selected food based on custom prompt answers
- Allows survey to ask clarifying questions about food, and possibly change to more appropriate food based on answers
- Applies to radio list, select, and yes / no prompt
- Configured within admin panel


packages/common/src/prompts/prompts.ts
```diff
@@ -226,6 +226,7 @@ const radioListPrompt = z.object({
   options: localeOptionList(),
   orientation: z.enum(radioOrientations),
   other: z.boolean(),
+  updateFood: z.boolean().default(false),
```
lukashroch, line 229: allow this to be extensible, something like `action: z.enum(['updateFood]).optional()`

packages/common/src/prompts/prompts.ts
```diff
@@ -234,6 +235,7 @@ const selectPrompt = z.object({
   component: z.literal('select-prompt'),
   options: localeOptionList({ limit: 2048 }),
   multiple: z.boolean(),
+  updateFood: z.boolean().default(false),
```
lukashroch, line 238: dtto

packages/common/src/prompts/prompts.ts
```diff
@@ -257,9 +259,13 @@ const timePickerPrompt = z.object({
 
 const yesNoPrompt = z.object({
   ...baseCustomPrompt.shape,
+  ...validatedPrompt.shape,
   component: z.literal('yes-no-prompt'),
   useFlag: z.boolean(),
   flag: z.string().optional(),
+  updateFood: z.boolean().default(false),
+  updateFoodOptionYes: z.string(),
```
lukashroch, line 267: IMO I think we should rather implement have ability to specific true/false values in generic way:
`trueValue: z.string().optional()`
`falseValue: z.string().optional()`
and fall bac kto booleans when not available

packages/common/src/types/common.ts
```diff
@@ -91,6 +91,7 @@ export function listOption<T extends z.ZodTypeAny = z.ZodString>(valueSchema?: T
     label: z.string().min(1).max(256),
     shortLabel: z.string().min(1).max(256).optional(),
     value: valueSchema ?? z.string().min(1).max(256),
+    updateFoodValue: z.string().min(1).max(256).optional(),
```
lukashroch, line 94: Do we need extra `updateFoodValue` field here? I mean could we simply use the `value`, is there a use-case for  `value` to differ from `updateFoodValue`

apps/survey/src/components/prompts/custom/yes-no-prompt.vue
```diff
@@ -45,10 +48,26 @@ const state = computed({
     emit('update:modelValue', value);
 
     if (typeof value === 'boolean')
-      action('next');
+      void submitPrompt(value);
```
lukashroch, line 51: redundant `void`

apps/survey/src/components/prompts/custom/update-food.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { foodsService } from '@intake24/survey/services';
+
+type ReplaceFoodFn = (data: { foodId: string; food: EncodedFood }) => void;
+
+type ReplaceFoodByCodeParams = {
+  food: EncodedFood | undefined;
+  foodCode: string | undefined;
+  localeId: string;
+  replaceFood: ReplaceFoodFn;
```
lukashroch, line 11: I would refrain from callbacks when possible and if design allows to keep the code/flow easier the read.
I think `replaceFoodByCode` can return replacement food entry, which can be passed into the survey store.
Also it might be write it as vue.js-style composable function as it is consumed by various prompts (i.e. components)

lukashroch (commented):
Thanks Miles, this looks useful.

My thoughts are that since this involves custom prompts, which we try keeping as generic as possible, I would implement the feature in more extensible fashion.

Instead of putting specific boolean property for `updateFood` only, I'd go for something like generic action consts that can be `updateFood` and code can act upon it. This way, it can be extended more easily in future. I don't mean to prematurely optimize, but more concerned with json schemas, i.e. rather to avoid migrating the scheme structure later, where we would have to have remove `updateFood` and turn it into something more generic later.

Also I have added couple of comments about the actual replacement values. I wonder if we need them or if we can use the prompt values. Is there any use-case to have these different?

@mechkg any thoughts on this?

packages/common/src/types/common.ts
miles-eastwood: `updateFoodValue` is the food code to be used as the replacement, whereas `value` is what is actually saved as the answer to the prompt. Researchers might be interested in the specific answer, and this cannot always be inferred from the final food code.

For example, suppose a radio-list prompt had three options, and only one option will update the food code. The other two options will both have `updateFoodValue` as `NO_UPDATE` (sentinel value), but they would both have distinct `value` fields to indicate the choice made by the respondent.

lukashroch:
@miles-eastwood OK I see re the updateValue, makes sense!

On second thought, I think we should use / extend existing `action` system we have for this.

Actions can be attached to any prompt (which are like buttons), but there are special cases where it's already not a button but like internal fn.

So I think the list option could have the `action` property in scheme, optional if there is applicable action to run.

There are defined here: https://github.com/intake24/intake24/blob/main/packages/common/src/prompts/actions.ts#L28

At the moment, `actionItem` has redundant props, only related to prompts for I think we should have something like
 
```ts
export const actionItem = z.object({
  type: z.enum(actionTypes),
  params: z.any(),
});

export const promptActionItem = actionItem.extends({
  text: localeTranslation,
  label: localeTranslation,
  color: z.string().nullable(),
  variant: z.enum(variants),
  icon: z.string().nullable(),
  layout: z.enum(layoutTypes).array(),
});
```

so for `updateFood` food we would have something like:
```json
{
  "type": "updateFood",
  "params": {
    "code": "SOME_FOOD_CODE"
  }
}
```

... `actionItem` should probably be better typed, probably a discriminated union to have params typed, it's a bit of TODO/WIP thing as params were recently used only for few things ...

Does that make sense?

miles-eastwood:
@lukashroch thanks for the feedback!

Could you clarify what you mean by:
> So I think the list option could have the `action` property in scheme, optional if there is applicable action to run.

Are you proposing that each individual option would have it's own action attached to it? Or would there be one `updateFood` action for the entire prompt, which is then configured via `params`?

Also, `yes-no-prompt` does not use `options-list`.

I can definitely try to re-implement this using the actions system, I am just not sure what the best way to connect the action (e.g which food code is used to update) to each individual option (or yes/no answer).

lukashroch:
I'd put action on each option in option list:
```ts
options: [
  {
    ...,
    action: action.optional()
  },
  ...
]
```

`yes-no-prompt` is kind of special so it would have to be on prompt level, something like:
```ts
trueAction: action.optional()
falseAction: action.optional()
```


We'd have to have a look how to best connect it to action system. At the moment, prompt actions bubble up like events from prompt component to top recall component and is handled that. Best would be extract the logic so it can be re-used for this, though there might be quite few dependencies that manipulate the recall flow, so might not be that straightforward. So at the moment as first draft, we could keep it separate / duplicated. And then next step could be to be revise that as it might be too much in one go. Best to keep the PR smaller.
This will still allow as to use same/similar action schemas across the survey scheme, we can improve the code logic later.

a bit OT, I was forced to rebase on main due to some unfinished worked that got into main to early, you might need to re-fetch the main and rebase your PR on top of that (when rebasing, drop the first commit - feat: food builders)

miles-eastwood:
Moved to PR #502

PR #499, opened 2026-06-22
Improve survey nutrient recalculation diagnostics and documentation

anguschiu1:
## Summary

Improve the safety and observability of survey nutrient recalculation, particularly for mixed-locale surveys and foods whose current mappings differ from their submitted food composition table (FCT).

## Changes

- Log missing `locale + food code` mappings with actionable context.
- Generate a downloadable CSV for foods skipped because they are unavailable in their submitted locale.
- Include skipped-food guidance in the completed job summary.
- Document recalculation modes, historical-data boundaries, mixed-locale behavior, and error handling.
- Add a nutrient data model guide with diagram support.
- Add integration coverage for cross-FCT recalculation scenarios.

## Testing

Added integration tests covering cross-FCT behavior for:

- `values-only` with and without `syncFields`
- `values-and-codes` with and without `syncFields`
- preservation or replacement of stored FCT references
- addition, removal, and zeroing of nutrient values

### DIFF
--- a/apps/docs/.vitepress/sidebar.ts
+++ b/apps/docs/.vitepress/sidebar.ts
@@ -234,6 +234,10 @@ export const sidebar = {
           link: '/admin/foods/nutrient-tables',
         },
         {
+          text: 'Nutrient data model',
+          link: '/admin/foods/nutrient-data-model',
+        },
+        {
           text: 'Nutrient types',
           link: '/admin/foods/nutrient-types',
         },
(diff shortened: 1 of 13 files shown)


apps/docs/.vitepress/sidebar.ts
```diff
@@ -328,6 +332,10 @@ export const sidebar = {
           text: 'Tasks',
           link: '/admin/surveys/tasks',
         },
+        {
```
lukashroch: I wouldn't put this one as separate thing, there as so many jobs, so would ne confusing to have random one there? Maybe have a link in `/admin/surveys/tasks` if not alread there?

lukashroch (changes requested):
LGTM, just one comment for docs.

apps/docs/.vitepress/sidebar.ts
anguschiu1: I think you're right, and `admin/surveys/tasks#surveynutrientsrecalculation` has links to that doc already. Give me five minutes to revise that!

PR #503, opened 2026-07-06
Enhance gtm events to track prompt changes 

anguschiu1:
## Summary

This PR improves GTM prompt tracking in the survey app by centralising prompt context in `gtmEvent.ts`.

It adds a `promptChanged` data-layer event whenever the active recall prompt changes, and includes current and previous prompt context on GTM custom events where available:

- `prompt_id`
- `prompt_component`
- `section`
- `previous_prompt_id`
- `previous_prompt_component`
- `previous_section`
- `action`

It also renames the analytics field from `scheme_prompts` to `section`, which better reflects the actual value being tracked.

## Why

Some custom GTM events did not include a useful `prompt_id`, making it difficult to map GA/GTM events back to the prompt the user was viewing. Prompt context is now tracked centrally, so standalone events can inherit the active prompt context without passing prompt IDs through component props manually.

## Validation

- `pnpm --filter @intake24/survey exec vitest run src/util/gtmEvent.spec.ts`
- `pnpm --filter @intake24/survey type-check`

### DIFF
--- a/apps/survey/src/components/recall/use-recall.ts
+++ b/apps/survey/src/components/recall/use-recall.ts
@@ -24,6 +24,7 @@ import {
   maybePushFallbackHistoryEntry,
   pushFullHistoryEntry,
 } from '@intake24/survey/stores/recall-history';
+import { recordPromptTransition } from '@intake24/survey/util';
 import { useI18n } from '@intake24/ui/i18n';
 
 export function useRecall() {
@@ -37,6 +38,11 @@ export function useRecall() {
   const currentPrompt = ref<PromptInstance | null>(null);
   const recallController = shallowRef<DynamicRecall | null>(null);
   const hideCurrentPrompt = ref(false);
+  const {
+    clearPromptTransition,
+    setCurrentPrompt,
+    setPromptTransitionAction,
+  } = createPromptTransitionController();
 
   const {
     hasFinished,
@@ -102,7 +108,8 @@ export function useRecall() {
 
     if (result === 'full') {
       hideCurrentPrompt.value = true;
-      currentPrompt.value = recallController.value?.getNextPrompt() ?? null;
+      setPromptTransitionAction('popstate');
+      setCurrentPrompt(recallController.value?.getNextPrompt() ?? null);
       hideCurrentPrompt.value = false;
     }
   };
@@ -112,7 +119,7 @@ export function useRecall() {
       return;
 
     // Prevent the currently active prompt from crashing if it expects a different selection type
-    currentPrompt.value = null;
+    setCurrentPrompt(null);
     survey.setSelection(newSelection);
   };
 
@@ -131,7 +138,7 @@ export function useRecall() {
       );
     }
 
-    currentPrompt.value = { section: promptSection, prompt };
+    setCurrentPrompt({ section: promptSection, prompt });
   };
 
   function showFoodPrompt(foodId: string, promptSection: MealSection, promptType: ComponentType) {
@@ -142,7 +149,7 @@ export function useRecall() {
     if (!prompt)
       throw new Error(`Survey scheme is missing required food prompt of type ${promptType}`);
 
-    currentPrompt.value = { section: promptSection, prompt };
+    setCurrentPrompt({ section: promptSection, prompt });
   };
 
   function showSurveyPrompt(promptSection: SurveyPromptSection, promptType: ComponentType) {
@@ -159,10 +166,12 @@ export function useRecall() {
       );
     }
 
-    currentPrompt.value = { section: promptSection, prompt };
+    setCurrentPrompt({ section: promptSection, prompt });
   };
 
   async function action(type: string, id?: string, params?: object) {
+    setPromptTransitionAction(type);
+
     switch (type) {
       case 'next':
         await next();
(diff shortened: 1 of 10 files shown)


apps/survey/src/util/gtmEvent.ts
```diff
@@ -4,6 +4,67 @@ import { useGtm } from '@gtm-support/vue-gtm';
...
+  prompt_component?: string;
+};
+
+type GtmPromptTrackingParams = PromptTransition & {
+  previous_prompt_id?: string;
+  previous_section?: string;
+  previous_prompt_component?: string;
+};
+
+let currentPrompt: PromptTransition | null = null;
```
lukashroch: Does these two need to be global? Could be declared in `recordPromptTransition` and then passed as params in `promptFields()` as args without holding global state.

apps/survey/src/components/recall/use-recall.ts
```diff
@@ -341,11 +358,14 @@ export function useRecall() {
 
   onMounted(async () => {
     addEventListener('popstate', onPopState);
+    promptTransitionAction = 'startRecall';
```
lukashroch: any refresh will call this? Not sure if it will actually reflect the "start recall" but maybe it's intentional/ok?

apps/survey/src/components/recall/use-recall.ts
```diff
@@ -324,6 +329,18 @@ export function useRecall() {
     });
   };
 
+  watch(currentPrompt, (promptInstance) => {
```
lukashroch: I wonder if we could get away to putting this into the `recallLog` store as we're sprinkling this GA stuff all around the code base. Recall log should already monitor any prompt change. But not sure if has all the params you capture here, so juts a thought if it similar logic could be closer together.

apps/survey/src/components/recall/use-recall.ts
anguschiu1: Yes, it is a mistake, I don't realise that will trigger in every refresh. As I have already installed "startRecall" event, this action should better name as "refreshPrompt"

lukashroch: maybe it's more a recall load, not really related to prompt. It will fire whenever the recall UI loads

apps/survey/src/components/recall/use-recall.ts
anguschiu1: After reviewing `recallLog`, I don't believe the parameters I need are as readily available in that context. Additionally, if I understand correctly, the `promptCheck` in `recallLog` can trigger multiple times before a single `getNextPrompt()` pass, and it logs "prompt change decision made by promptManager" rather than the actual "prompt change" event.

For these reasons, I believe it's better to keep the GTM event trigger in its current location, where it fires when the actual prompt change occurs.

I've refactored the code to replace the watcher with an explicit `setCurrentPrompt` method to make the intention clearer. I'd appreciate your thoughts on this approach.

apps/survey/src/util/gtmEvent.ts
anguschiu1: We need some module-level context so later `sendGtmEvent` calls can inherit the active prompt. I’ve wrapped it as an explicit `promptContext` object and passed that into `promptFields` to make the state ownership clearer.

apps/survey/src/components/recall/use-recall.ts
anguschiu1: Yes, I updated the `action` of this event as `loadRecallUI`.

apps/survey/src/components/recall/use-recall.ts
```diff
@@ -35,6 +36,7 @@ export function useRecall() {
   const { i18n: { locale } } = useI18n();
 
   const currentPrompt = ref<PromptInstance | null>(null);
+  let promptTransitionAction = 'not mapped';
```
lukashroch: Similar to before, wouldn't this be better for this to be arg of `setCurrentPrompt`. Being this as global var, this could lead to subtle side-effects, like wrong transition attached to GA event. Since this is not being used for anything else, it doesn't need to be global and `setCurrentPrompt` can explicitly set the transition action with more predictable behaviour?

lukashroch (changes requested):

apps/survey/src/components/recall/use-recall.ts
lukashroch: Makes sense with recallLog, I didn't realize it iterate over and records everything.

apps/survey/src/components/recall/use-recall.ts
anguschiu1: I’ve scoped `promptTransitionAction` into a small prompt transition controller, alongside the `setCurrentPrompt()` logic that consumes it.

I considered passing the transition action directly through `setCurrentPrompt()`, but doing that cleanly would require threading a parameter only conerned by GTM tracking through `next()`, `nextPrompt()`, `recallAction()`, and the meal/food prompt helpers, something like:

```
async function next(transitionAction = 'next') {...}
```

Those functions are part of the main recall flow, so I wanted to avoid increasing the blast radius, keeps the existing recall flow signatures unchanged, and coupling the navigation helpers more tightly to GTM tracking.

Now the state being private within the recording logic and no longer loose in `useRecall()`, and the read/write/reset lifecycle is grouped in one place.

Please let me know your thoughts to make it more clean and succinct.

lukashroch (approved):

PR #504, opened 2026-07-13
fix(food-index-ranking): normalize sorting algorithm naming

miles-eastwood:
**Problem**
A naming mismatch (`rankingAlgorithm` vs `sortingAlgorithm`) causes the api to always return the default (`'popularity'`), even if a different algorithm is stored in the DB.

**Fix**
Replace all instances of `rankingAlgorithm` with `sortingAlgorithm`, which is the name used in the DB.

Also removed type definition `RankingAlgorithm` that appears to be unused.


lukashroch (approved):
Nice, thanks! That's a subtle one with all the optional params.

PR #505, opened 2026-07-27
feat(prompt-list): prompt editor subsections

miles-eastwood:
- Adds subsections to the survey-scheme prompts editor. This allows for easier management when schemes have a large number of prompts.
- Subsections can be created, deleted, and renamed.
- Subsections are draggable and can be rearranged. Prompts can also be dragged between subsections (within the same section)

### DIFF
--- a/packages/common/src/surveys/scheme.ts
+++ b/packages/common/src/surveys/scheme.ts
@@ -52,6 +52,25 @@ export type FoodSection = (typeof foodSections)[number];
 export const promptSections = [...surveySections, ...mealSections] as const;
 export type PromptSection = (typeof promptSections)[number];
 
+export const promptSubsectionLayout = z.object({
+  id: z.string(),
+  size: z.coerce.number().int().min(0),
+  expanded: z.boolean(),
+  name: z.string(),
+});
+export type PromptSubsectionLayout = z.infer<typeof promptSubsectionLayout>;
+
+export const promptSubsectionLayouts = z.object({
+  preMeals: promptSubsectionLayout.array(),
+  preFoods: promptSubsectionLayout.array(),
+  foods: promptSubsectionLayout.array(),
+  postFoods: promptSubsectionLayout.array(),
+  foodsDeferred: promptSubsectionLayout.array(),
+  postMeals: promptSubsectionLayout.array(),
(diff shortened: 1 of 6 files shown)


packages/common/src/surveys/scheme.ts
```diff
@@ -52,6 +52,25 @@ export type FoodSection = (typeof foodSections)[number];
...
 
+export const promptSubsectionLayout = z.object({
+  id: z.string(),
+  size: z.coerce.number().int().min(0),
+  expanded: z.boolean(),
+  name: z.string(),
+});
+export type PromptSubsectionLayout = z.infer<typeof promptSubsectionLayout>;
+
+export const promptSubsectionLayouts = z.object({
```
lukashroch, line 63: maybe use `z.record` of keys as value is the same?

packages/common/src/surveys/scheme.ts
```diff
@@ -77,6 +96,9 @@ export const recallPrompts = z.object({
   }),
   postMeals: singlePrompt.array(),
   submission: singlePrompt.array(),
+  ui: z.object({
+    subsectionLayouts: promptSubsectionLayouts,
```
lukashroch, line 100: would use just `subsections` to keep the scheme keys concise

lukashroch (changes requested):
Looks good, UI seems more organized. Let me put it into DEV branch and discuss with tech team as it changes the scheme editor quite a bit, so would be good to have broader input.

Very few code comments and maybe few UI changes thoughts:
- "add prompt" in section is redundant, consider removing since user should use the one in subsection? It's already a bit cluttered with buttons there
- add missing spacing between delete / expand in subsection title controls

@mechkg could you have a look / your thoughts?

lukashroch:
@miles-eastwood apologies it's taking so long, people in teams are quite on leaves.

I have pulled this into our spring to review and had positive feedback. I think we should be able to pull this in. If you could have a look into those few comments and address, I'll give it a second read and merge then if all OK.

miles-eastwood:
@lukashroch Sorry for not getting to those issues, I was off for a few weeks. Thanks for fixing / merging it in!

PR #506, opened 2026-07-29
Accessible image-based portion selection

miles-eastwood:
Currently, image-based portion selection in intake24 is not accessible to those using a screenreader. This update fixes that by adding keyboard navigability and `aria-label`s to each image, which will be read by a screenreader.

`aria-label`s are generated using the existing label system. These labels are now fetched unconditionally. `labelsEnabled` still controls label visibility.

### DIFF
--- a/apps/survey/src/components/prompts/portion/PizzaPrompt.vue
+++ b/apps/survey/src/components/prompts/portion/PizzaPrompt.vue
@@ -205,8 +205,6 @@ const imageMapIds = computed(() => ({
 }));
 
 const labelsEnabled = computed(() => !!(props.prompt.imageMap.labels && !!parameters.value.labels));
-console.log(props.prompt.imageMap.labels);
-console.log(parameters.value);
 const imageMapLabels = computed(() => {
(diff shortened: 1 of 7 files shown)


apps/survey/src/components/prompts/portion/PizzaPrompt.vue
```diff
@@ -201,16 +204,10 @@ const imageMapIds = computed(() => ({
   slice: sliceImageMapId.value,
 }));
 
-const labelsEnabled = computed(() => props.prompt.imageMap.labels && !!parameters.value.labels);
+const labelsEnabled = computed(() => !!(props.prompt.imageMap.labels && !!parameters.value.labels));
+console.log(props.prompt.imageMap.labels);
```
lukashroch: redundant logs

lukashroch (changes requested):
LGTM

only redundant console logs

apps/survey/src/components/prompts/portion/PizzaPrompt.vue
miles-eastwood: removed.

PR #507, opened 2026-08-10
fix(same-as-before): set props to null instead of undefined

miles-eastwood:
Setting `portionSizeMethodIndex` and `portionSize` to `undefined` results in a runtime error, as these values should be assigned `null` if they are unknown (and `undefined` values pass `null` checks).

### DIFF
--- a/apps/survey/src/components/handlers/standard/SameAsBeforePromptHandler.vue
+++ b/apps/survey/src/components/handlers/standard/SameAsBeforePromptHandler.vue
@@ -42,7 +42,7 @@ const updatedSabFood = {
 function onSabOptionsUpdate(sabOptions: Record<string, boolean>): void {
   console.debug('Received sabOptions from child:', sabOptions);
   if (!sabOptions.portionSize && updatedSabFood.food.portionSize) {
-    Object.assign(updatedSabFood.food, { portionSizeMethodIndex: undefined, portionSize: undefined });
+    Object.assign(updatedSabFood.food, { portionSizeMethodIndex: null, portionSize: null });
     updatedSabFood.food.flags = updatedSabFood.food.flags.filter(flag => !['portion-size-option-complete', 'portion-size-method-complete'].includes(flag));
     console.debug(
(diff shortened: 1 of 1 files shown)


apps/survey/src/components/handlers/standard/SameAsBeforePromptHandler.vue
```diff
@@ -42,7 +42,7 @@ const updatedSabFood = {
 function onSabOptionsUpdate(sabOptions: Record<string, boolean>): void {
   console.debug('Received sabOptions from child:', sabOptions);
   if (!sabOptions.portionSize && updatedSabFood.food.portionSize) {
-    Object.assign(updatedSabFood.food, { portionSizeMethodIndex: undefined, portionSize: undefined });
+    Object.assign(updatedSabFood.food, { portionSizeMethodIndex: null, portionSize: null });
```
lukashroch: Nice catch. IMO we should improve this a bit more as this was missed due to not properly typed `updatedSabFood` var a bit above.

`const updatedSabFood` should be typed as `const updatedSabFood: SameAsBeforeItem`,

then this will also trigger similar errors on `updatedSabFood` var assignment few lines above for
```ts
portionSize: sabFood?.food?.portionSize ? { ...sabFood.food.portionSize } : undefined,
    portionSizeMethodIndex: sabFood?.food?.portionSizeMethodIndex ?? undefined,
```
... which should be nulls to

We could probably remove Object.assign with simple assignment for type-safety sake.

Would you be OK to extend this?

lukashroch (changes requested):

apps/survey/src/components/handlers/standard/SameAsBeforePromptHandler.vue
miles-eastwood: `useSameAsBefore().getItem(survey.localeId, code)` returns type `SameAsBeforeItem | undefined`, so I amended `updatedSabFood` to match that.

PR #514, opened 2026-08-28
Improve prompt tracking

anguschiu1:
## Summary

This enhancement seeks to follow up the UX monitoring [Jira ticket](https://intake24.atlassian.net/jira/software/c/projects/V4/boards/5?selectedIssue=V4-1805)

- Add GTM tracking for prompt choices, history direction, carousel navigation, feedback/follow-up links, and persistent meal-list (quick-list) actions.
- Preserve prompt context while distinguishing meal-list deletions from prompt-driven deletions.
- Fix food-search ordering so the first typed search is recorded as `1`.
- Add focused coverage for new tracking and history behaviour.

## Testing

- `pnpm --filter @intake24/survey test:unit --run src/util/gtmEvent.spec.ts src/stores/recall-history.spec.ts src/components/layouts/meal-list/use-meal-list.spec.ts src/components/elements/FoodBrowser.spec.ts` — 4 files, 35 tests passed.

## Follow up
GTM container will need to update separately, once the changes are applied to production and emit events.


apps/survey/src/components/recall/use-recall.ts
```diff
@@ -428,11 +438,12 @@ export function useRecall() {
         prompt_id: promptInstance.prompt.id,
         section: promptInstance.section,
         prompt_component: promptInstance.prompt.component,
-      });
+      }, option === undefined ? undefined : { selected_option: option });
     }
```
mechkg, line 468: This mechanism for saving `selectedOption` will likely break when prompt selection changes because prompt transition on selection injects a `setCurrentPrompt(null)` call (see line 127 in the same file). This call will clear `selectedOption` prematurely and it will be lost when the actual next prompt is selected.

`selectedOption = undefined` should be after `if (!promptInstance)` check.

apps/survey/src/util/gtmEvent.ts
```diff
@@ -133,3 +144,38 @@ export function sendGtmEvent(params: GtmEventParams): void {
...
+
+export function sendMealListEvent(event: MealListAction): void {
+  sendGtmEvent({ event, action: mealListActions[event] });
+}
+
+export function sendDeletionEvent(
+  event: 'deleteFood' | 'deleteMeal',
+): void {
+  sendGtmEvent({ event, action: event });
+}
```
mechkg, line 183: Deletion event does not include information like meal/food name, is that intentional?

apps/survey/src/components/prompts/standard/RedirectPrompt.vue
```diff
@@ -37,6 +38,7 @@
             :href="followUpUrl"
             :target="prompt.target"
             :title="promptI18n.goTo"
+            @click="sendFeedbackLinkEvent('feedbackLinkClicked', 'follow_up')"
           >
```
mechkg, line 42: AI flags this as a potentially lost GTM event, because sendFeedbackLinkEvent only adds the event to an array that is then processed asynchronously, but clicking the link will also unload the current page immediately so the outcome is unpredictable.

Not sure if this case is handled by GTM, worth looking into.

mechkg (changes requested):
A couple of issues - please see comments. The main one is `selectedOption` being cleared prematurely.

apps/survey/src/components/recall/use-recall.ts
anguschiu1: Thank you for point it out! I fixed it already. 

And then I realise that the option tracked should be labelled as `previous_selected_option`because it will log into the immediate next gtm event, instead of the event that asked for selection, i.e.

Some options-prompt or yes-no-prompt -> custom event fired 
User selected "Yes" -> logged
Prompt changed -> custom event fired, along with the variable `previous_selected_option` to tell how to reach this prompt

apps/survey/src/util/gtmEvent.ts
anguschiu1: I presumed GTM can provide the context like addFood and selectFood event, but it is not the case for deleteFood/ deleteMeal. 
So it's a bug and fixed, and standardise them to emit food/meal along the event. Thank you very much!

apps/survey/src/components/prompts/standard/RedirectPrompt.vue
anguschiu1: Confirmed it is not an issue as GTM will track and generate an link click event, even the link is to pointing other domain like "https://redcap.mrc-epid.cam.ac.uk/surveys/", so it is fine.

But I changed the variable name to improve clarify.

mechkg (approved):

PR #515, opened 2026-09-02
feat(admin): V4-1752 add capability to export NDB mapping and NDB data CSVs

anguschiu1:
## Summary
- Add two new tasks `NutrientTableDataExport` and `NutrientTableMappingExport`, to export nutrient-table NDB mapping and data CSVs compatible with the existing import tasks
- Generated NDB data CSV will contain column headers, but become header-less if `rowOffset` is 0.
- report exported record and CSV-column counts for NDB data and mapping exports
- clarify the nutrient-table record reference ID used by locale food nutrient mappings
- consolidate CSV file check into a util function `requireCsvUploadPath`

## Verification
- NDB data export/import round-trip unit test and integration test (i.e. export NDB can be used for import)
- NDB mapping export/import unit test and round-trip integration tests
- API typecheck and lint


apps/api/src/jobs/nutrient-tables/nutrient-table-mapping-export.ts
```diff
@@ -0,0 +1,86 @@
...
+      where: { nutrientTableId },
+      attributes: ['nutrientTypeId', 'columnOffset'],
+      include: [{ association: 'nutrientType', attributes: ['description'] }],
+      order: [['columnOffset', 'ASC'], ['nutrientTypeId', 'ASC']],
+    });
+
+    const timestamp = format(new Date(), 'yyyyMMdd-HHmmss');
+    const filename = `intake24-${this.name}-${nutrientTableId}-${timestamp}.csv`;
+    const output = createWriteStream(path.resolve(this.fsConfig.local.downloads, filename), { encoding: 'utf-8', flags: 'w+' });
+    const records = Readable.from(mappings.map(mapping => ({
```
lukashroch: I think this beats the purpose to use streaming, it's all loaded into the memory by above query. I'd write it with `kysely` with stream support. Similarly second job too?

I think you can lift and adjust some usage already from the source, e.g. resources exports are done in this way (https://github.com/intake24/intake24/blob/main/apps/api/src/jobs/system/resources/as-served-sets.ts#L15)

lukashroch (changes requested):

apps/api/src/jobs/nutrient-tables/nutrient-table-mapping-export.ts
anguschiu1: Thanks for the advice! Using `stream()` from Kysely and add progress icon state change to NDB data export.

apps/api/src/jobs/nutrient-tables/nutrient-table-data-export.ts
```diff
@@ -0,0 +1,182 @@
...
+    const { total } = await this.kyselyDb.foods
+      .selectFrom('nutrientTableRecords')
+      .select(({ fn }) => [fn.count<number>('id').as('total')])
+      .where('nutrientTableId', '=', nutrientTableId)
+      .executeTakeFirstOrThrow();
+    this.initProgress(Number(total));
+
+    const fields = new Map(csvMappingFields.map(mapping => [mapping.fieldName, mapping.columnOffset]));
+    const nutrients = new Map(csvMappingNutrients.map(mapping => [mapping.nutrientTypeId, mapping.columnOffset]));
+    const cursor = this.kyselyDb.foods
```
mechkg, line 98: This query can potentially explode because of the double join, each record will produce number of fields x number of nutrients rows this way.  

Not critical for our data but still dangerous. It's safer to do it as three separate streams ordered by `nutrientTableRecordId` (they should be within a transaction to guard against simultaneous writes).

You can probably also do it with an aggregation function (like json_agg) but it's SQL dialect-specific, I think three separate queries is cleaner.

mechkg (changes requested):

apps/api/src/jobs/nutrient-tables/nutrient-table-data-export.ts
```diff
@@ -0,0 +1,182 @@
...
+      csvMapping.idColumnOffset,
+      csvMapping.descriptionColumnOffset,
+      csvMapping.localDescriptionColumnOffset ?? 0,
+      ...csvMappingFields.map(({ columnOffset }) => columnOffset),
+      ...csvMappingNutrients.map(({ columnOffset }) => columnOffset),
+    );
+    const header = Array.from<string>({ length: maxOffset + 1 }).fill('');
+    header[csvMapping.idColumnOffset] = 'NDB food ID (FCT record ID)';
+    header[csvMapping.descriptionColumnOffset] = 'NDB food description';
+    if (csvMapping.localDescriptionColumnOffset)
```
mechkg: Is the intent here (csvMapping.localDescriptionColumnOffset !== undefined)? 

It could drop valid `0` offsets otherwise

anguschiu1: Good catch, thanks — 0 is a valid column offset and the truthiness check incorrectly treated it as unset, i.e. the bug occurs when a user explicitly enters A, instead of leaves it blank, to `Local description column`.
I found that import task also has this problem and I fixed it as well.

apps/api/src/jobs/nutrient-tables/nutrient-table-data-export.ts
anguschiu1: Thank you for pointing out — the original double LEFT JOIN could produce a fields × nutrients x nutrient-table record record set that reach 2M rows.

A bit of digging, I replaced it with one ordered UNION ALL stream of record, field, and nutrient fragments, and the existing row builder groups them by record ID into one CSV row, including records with no fields or nutrients.

One UNION ALL query also keeps a consistent snapshot while streaming; three concurrent Kysely streams cannot safely share one transaction connection.

Kindly advise if it is the right approach, or I should load the results of 3 queries into memory and process.

mechkg: You're right that three concurrent streams wouldn't work in our current setup. I tested it and it stalls if they are in the same transaction and you try to interleave them. It also stalls other plain (non-stream) queries in the same transaction.

The union based query avoids the cross product, but it is still very expensive. I ran it through `EXPLAIN` on a real database snapshot and it scanned *all* of the nutrient table rows in the database, then sorted them in memory (and they couldn't fit in memory).

What seems to be the most predictable given these limitations is a simple batched loop within a transaction, without .stream(), like this:

```
select ... from nutrient_table_records
where nutrient_table_id = 'NDNS'
order by nutrient_table_record_id
limit {BATCH_SIZE};
```

and then:
```
select ... from nutrient_table_records
where nutrient_table_id = ...
  and nutrient_table_record_id > {LAST_ID_FROM_PREVIOUS_BATCH}
order by nutrient_table_record_id
limit {BATCH_SIZE};
```

a where clause with `> last_id`  is more efficient than `offset` for batching if there is an appropriate index on that column. You stop when it returns fewer than BATCH_SIZE records.

and then you just do a simple `where nutrient_table_record_id in (...)` on `nutrient_table_record_nutrients` and `nutrient_table_record_fields` for every batch.

